PR #2, opened 2023-10-21
Update the .scripts submodule too

tianon:
We recently had an issue where BuildKit was being tried for Windows builds, which is fixed (in https://github.com/docker-library/meta-scripts/commit/1531fc9fc8e35480ed46e5c900dd24c4937580c3), but wasn't applied because the scripts submodule wasn't being updated (and really should be, per the `TODO` comment).


whalelines (approved):
LGTM

PR #3, opened 2023-11-10
Cache known architectures to reduce calls to registry from bashbrew

LaurentGoderre:


builds.sh
```diff
@@ -130,11 +140,18 @@ for sourceId; do
 				end
 			')"
 
+			if [ -n "imageArchResolved["$lookup"]:+x" ]; then
```
whalelines, line 143: What is this line doing? From the following lines, it looks like you want to check if the value for the key `$lookup` of the associative array `imageArchResolved` has no value since when true you calculate the value and cache it and when false you just use the cached value. If this is the case, I would expect such test to look like this:

```sh
if [ -z "${imageArchResolved[$lookup]:+x}" ]; then
```

Note switching from `-n` to `-z` and and wrapping the evaluation of the associative array key and `:+` in `${}`.

As it stands now, this conditional will either always be true or error, at least if my internal bash parser is correct. It will be true if `$lookup` is zero or one "word" and will error if `$lookup` is two or more words. Since you are not dereferencing the `imageArchResolved` variable, i.e., as `${imageArchResolved[KEY]}`, the `imageArchResolved[` is just that static string, then outside the double quotes you expand `$lookup`, and then end with the static string `]:+x`, which, again, is not inside `${}`, so it just evaluates as itself. Since `$lookup` is not quoted, if it contains whitespace it will expand to more than one word and bash will complain that `-n` only takes a single argument.

Not for this PR, but it is often safer and clearer to use `[[ ]]` in bash.

whalelines (commented):
Please see my comment about evaluation of the conditional around caching. It seems like the code as it is now would never use the cache, so it is curious how it is so much faster. But perhaps I am missing something. Is the output of the script the same before and after these changes?

builds.sh
whalelines: To make things more concrete:

```sh
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; if [ -n "imageArchResolved["$lookup"]:+x" ]; then echo "fetch & cache"; else echo "use cache"; fi'
fetch & cache
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; imageArchResolved+=([$lookup]=386); if [ -n "imageArchResolved["$lookup"]:+x" ]; then echo "fetch & cache"; else echo "use cache"; fi'
fetch & cache
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; if [ -n "${imageArchResolved[$lookup]:+x}" ]; then echo "fetch & cache"; else echo "use cache"; fi'
use cache
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; imageArchResolved+=([$lookup]=386); if [ -n "${imageArchResolved[$lookup]:+x}" ]; then echo "fetch & cache"; else echo "use cache"; fi'
fetch & cache
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; if [ -z "${imageArchResolved[$lookup]:+x}" ]; then echo "fetch & cache"; else echo "use cache"; fi'
fetch & cache
$ bash -c 'declare -A imageArchResolved; lookup=debian:latest; imageArchResolved+=([$lookup]=386); if [ -z "${imageArchResolved[$lookup]:+x}" ]; then echo "fetch & cache"; else echo "use cache"; fi'
use cache
```

LaurentGoderre: I had fixed that but pushed to the wrong repo!

LaurentGoderre:
The `${...}` was missing in this PR (I pushed to the wrong place. As for using `-n` or `-z`, I reused the logic structure of the existing condition here: https://github.com/docker-library/meta-scripts/blob/main/builds.sh#L134

builds.sh
```diff
@@ -130,11 +141,18 @@ for sourceId; do
 				end
 			')"
 
+			if [ -n "${imageArchResolved["$lookup"]:+x}" ]; then
```
whalelines, line 144: You've kept the `-n`, but swapped the bodies of the `if` and the `else`, which is the same as using `-z` and not swapping the bodies.

whalelines (commented):
So this present version has:

1.  been tested?
2.  is faster?
3.  produces the same output as the released version?

LaurentGoderre:
I was limited last week because i hit the daily quota. I had one last minor issue and not it runs ansd is faster (20 sec now). The output is different in some places though so I am checking that now.

builds.sh
```diff
@@ -130,11 +147,18 @@ for sourceId; do
 				end
 			')"
 
+			if [ -v "imageArchResolved["$lookup"]" ]; then
```
whalelines, line 150:
```suggestion
			if [ -n "${imageArchResolved[$lookup]+x}" ]; then
```

It would be clearer/cleaner/more idiomatic to use the above formulation. Plus, while `$lookup` should never have a space in it, it is also safer/future proof.

What I'm not clear on is whether you should use `:+` or `+`. The latter (the one recommended in the suggestion) maps more closely to the result of `-v`. The question is whether it is legitimate for resolved architectures for an image to be null?

builds.sh
```diff
@@ -160,11 +184,20 @@ for sourceId; do
...
 			buildId="$(sha256sum <<<"$buildIdJson" | cut -d' ' -f1)" # see notes above (where buildIdParts is first defined)
 			printf >&2 '%s\n' "$buildId"
+
 			img="$BASHBREW_STAGING_TEMPLATE"
 			img="${img//BUILD/$buildId}"
 			[ "$img" != "$BASHBREW_STAGING_TEMPLATE" ] # BUILD is required, for proper uniqueness (ARCH is optional)
 			img="${img//ARCH/$arch}"
-			resolved="$(_resolve "$img" "$arch")"
+
+			if [ -v "imageArchResolved["$img"]" ]; then
```
whalelines, line 193:
```suggestion
			if [ -n "${imageArchResolved[$img]+x}" ]; then
```

builds.sh
```diff
@@ -130,11 +147,18 @@ for sourceId; do
 				end
 			')"
 
+			if [ -v "imageArchResolved["$lookup"]" ]; then
+				remoteArches="${imageArchResolved["$lookup"]}"
```
whalelines, line 151:
```suggestion
				remoteArches=${imageArchResolved[$lookup]}
```

There is no need to quote any of this. Word splitting is not done on the right side of assignments.

builds.sh
```diff
@@ -130,11 +147,18 @@ for sourceId; do
...
+			else
+				remoteArches=$(_resolveRemoteArch "$lookup")
+				imageArchResolved["$lookup"]="$remoteArches"
```
whalelines, line 154:
```suggestion
				imageArchResolved[$lookup]=$remoteArches
```

builds.sh
```diff
@@ -19,13 +19,26 @@ shell="$(jq -r '
...
 eval "$shell"
 
+_resolveRemoteArch() {
+	local img="$1"; shift
+
+	local arches
+	if ! arches="$(bashbrew remote arches --json "$img" 2>/dev/null)"; then # TODO somehow differentiate errors like 404 / 403, "insufficient_scope" from other bashbrew errors here so we can stop eating stderr
+		return
+	fi
+
```
whalelines, line 29: Perhaps we should be caching here rather than repeating the code twice in the two places where this function is called.

builds.sh
```diff
@@ -19,13 +19,26 @@ shell="$(jq -r '
...
 # resolve an image reference to an architecture-specific imageId (digest of the image manifest)
 _resolve() {
 	local img="$1"; shift
 	local arch="$1"; shift
+	local bashbrewRemoteArches=
 
-	local bashbrewRemoteArches
-	if ! bashbrewRemoteArches="$(bashbrew remote arches --json "$img" 2>/dev/null)"; then # TODO somehow differentiate errors like 404 / 403, "insufficient_scope" from other bashbrew errors here so we can stop eating stderr
+	if [ -n "${1:+x}" ]; then
```
whalelines, line 39: This looks like the resolved architectures could be null/empty.

builds.sh
whalelines: It looks like it can be empty, so just `+` is probably better.

builds.sh
```diff
@@ -160,11 +184,20 @@ for sourceId; do
...
 			printf >&2 '%s\n' "$buildId"
+
 			img="$BASHBREW_STAGING_TEMPLATE"
 			img="${img//BUILD/$buildId}"
 			[ "$img" != "$BASHBREW_STAGING_TEMPLATE" ] # BUILD is required, for proper uniqueness (ARCH is optional)
 			img="${img//ARCH/$arch}"
-			resolved="$(_resolve "$img" "$arch")"
+
+			if [ -v "imageArchResolved["$img"]" ]; then
+				remoteArches="${imageArchResolved["$img"]}"
```
whalelines, line 194:
```suggestion
				remoteArches=${imageArchResolved[$img]}
```

builds.sh
```diff
@@ -160,11 +184,20 @@ for sourceId; do
...
+			else
+				remoteArches=$(_resolveRemoteArch "$img")
+				imageArchResolved["$img"]="$remoteArches"
```
whalelines, line 197:
```suggestion
				imageArchResolved[$img]=$remoteArches
```

whalelines (commented):
A few adjustments and clarifications.

builds.sh
tianon: The quoting here doesn't cause any change in behavior and ensures we keep up on our quoting discipline more consistently, so I'd prefer to keep these types of assignments quoted.

builds.sh
tianon: Same comment as above, I prefer this one as is in the PR.

builds.sh
tianon: Ditto

builds.sh
```diff
@@ -56,6 +69,10 @@ _resolve() {
 	' || exit 1
 }
 
+declare -A imageArchResolved=(
+	#["image"]="arch"
```
tianon, line 73: This should make the format more clear, something like:
```suggestion
	#["image"]="{...}" # JSON result of bashbrew remote arches
```

builds.sh
```diff
@@ -19,13 +19,26 @@ shell="$(jq -r '
 ' "$json")"
 eval "$shell"
 
+_resolveRemoteArch() {
+	local img="$1"; shift
+
+	local arches
+	if ! arches="$(bashbrew remote arches --json "$img" 2>/dev/null)"; then # TODO somehow differentiate errors like 404 / 403, "insufficient_scope" from other bashbrew errors here so we can stop eating stderr
+		return
```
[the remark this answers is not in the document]
tianon: (this would also make the empty vs not set concerns go away and we could use the `:` forms consistently and be less likely to break the logic in future changes)

builds.sh
LaurentGoderre: That was what I tried at first but there's a lot of subshells going in on and the cache gets lost if we put it here. That's why it ended where it is now.

builds.sh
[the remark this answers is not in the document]
whalelines: +1 to both

builds.sh
whalelines: Ah yes, caching a returned value is tricky in shell.

builds.sh
whalelines: In that case, should the cached "null" value be `{}`?

builds.sh
whalelines: My concern here is how Bash deals with the nested double quotes. Does it _do the right thing™_? I know it does for sub-shells and it does not for single quotes, but I'm not sure about inside `${}` and it's difficult to tell since the square brackets effectively behave as quotes, i.e., the index can have spaces and this is perfectly safe: `${a[b c]}`. Interesting to note, I've seen constructs like this elsewhere in the code base: `v='w'$x'y'`, i.e., they leverage the fact that the right-hand side of the assignment operator do not undergo word splitting. So perhaps `"${imageArchResolved[$lookup]}"` is the cleanest option?

builds.sh
whalelines:
```suggestion
			if [ -n "${imageArchResolved[$img]:+x}" ]; then
```

If we also explicitly cache the null/empty/`{}` value.

builds.sh
whalelines: But `:+` if we cache the empty value.

LaurentGoderre: Done

builds.sh
LaurentGoderre: Done

builds.sh
LaurentGoderre: Yes

builds.sh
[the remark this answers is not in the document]
LaurentGoderre: Not sure why it doesn't update here but I added `printf null`

builds.sh
tianon: If there's anywhere I haven't consistently written `v='w'$x'y'` as `v='w'"$x"'y'`, we should fix it.

Each of `foo[$bar]=...` vs `foo["$bar"]=...` and `"${foo[$bar]}"` vs `"${foo["$bar"]}"` do indeed work interchangeably, and do indeed work correctly (even when `$bar` contains multiple successive whitespaces):

```console
$ declare -A foo=()
$ bar='a   b' # note the three spaces
$ foo[$bar]=xyz
$ foo["$bar"]=abc
$ echo "${!foo[@]}"
a   b
$ echo "${foo[$bar]}"
abc
$ echo "${foo["$bar"]}"
abc
```

builds.sh
tianon: I think `null` is fine if we're going to _pre_-test against it -- if we want to use a different value, it should be something like `{"arches":{}}` that we wouldn't _have_ to pre-test against (ie, that the `jq` would just "do the right thing" with), but IMO `null` is more clear as a sentinel that the lookup had some kind of problem (such as the image not existing).

builds.sh
```diff
@@ -19,13 +19,27 @@ shell="$(jq -r '
...
+	fi
+
+	echo "$arches"
+}
+
 # resolve an image reference to an architecture-specific imageId (digest of the image manifest)
 _resolve() {
 	local img="$1"; shift
 	local arch="$1"; shift
+	local bashbrewRemoteArches=
```
tianon, line 38: Now that we're always passing a value (either a JSON object or the JSON null value), this can be reverted and the conditional can be simply `if [ "$bashbrewRemoteArches" = 'null' ]; then printf 'null'; return; fi`

builds.sh
```diff
@@ -130,11 +148,18 @@ for sourceId; do
 				end
 			')"
 
+			if [ -n "${imageArchResolved[$lookup]:+x}" ]; then
+				remoteArches="${imageArchResolved["$lookup"]}"
+			else
+				remoteArches=$(_resolveRemoteArch "$lookup")
```
tianon, line 154:
```suggestion
				remoteArches="$(_resolveRemoteArch "$lookup")"
```

(for consistency)

builds.sh
```diff
@@ -160,11 +185,20 @@ for sourceId; do
...
 			img="$BASHBREW_STAGING_TEMPLATE"
 			img="${img//BUILD/$buildId}"
 			[ "$img" != "$BASHBREW_STAGING_TEMPLATE" ] # BUILD is required, for proper uniqueness (ARCH is optional)
 			img="${img//ARCH/$arch}"
-			resolved="$(_resolve "$img" "$arch")"
+
+			if [ -n "${imageArchResolved[$img]:+x}" ]; then
+				remoteArches="${imageArchResolved["$img"]}"
+			else
+				remoteArches=$(_resolveRemoteArch "$img")
```
tianon, line 197:
```suggestion
				remoteArches="$(_resolveRemoteArch "$img")"
```

builds.sh
whalelines: Ah, `v='w'"$x"'y'` is probably what I was thinking of. Sorry, didn't mean to impugn your work.

tianon:
Tested this (including all my suggestions) with my personal images and can confirm that it is a slight speedup, although it's not perfectly representative of all of DOI (just a bit more than our current subset here):

```console
$ time ./builds.sh sources.json &> /dev/null

real	2m37.893s
user	1m2.352s
sys	0m8.182s

$ time ./new-builds.sh sources.json &> /dev/null

real	1m52.489s
user	0m53.636s
sys	0m5.700s
```

whalelines:
Thanks for testing this out (even though you shouldn't be!).

I spoke with @yosifkit and @LaurentGoderre about what might be a larger issue: that running these scripts on an EC2 instance is about twice a slow as running it locally. This despite the EC2 instance, a t3a.2xlarge, having 8 vCPU (with plenty of CPU credits) and 32 GB of RAM.

builds.sh
```diff
@@ -19,15 +19,27 @@ shell="$(jq -r '
...
 	local arch="$1"; shift
+	local bashbrewRemoteArches="$1"; shift
 
-	local bashbrewRemoteArches
-	if ! bashbrewRemoteArches="$(bashbrew remote arches --json "$img" 2>/dev/null)"; then # TODO somehow differentiate errors like 404 / 403, "insufficient_scope" from other bashbrew errors here so we can stop eating stderr
-		printf 'null'
-		return
+	if [ "$bashbrewRemoteArches" = 'null' ]; then
+		printf 'null';
+		return;
```
tianon, line 42: This must be my bad for giving you a one-liner example: :sweat_smile:

```suggestion
		printf 'null'
		return
```

LaurentGoderre: It's also a habit for me doing image commands one-liners

whalelines: If we're removing unnecessary characters, there is no need to quote a one-word literal.

LaurentGoderre: I'm just going on consistency with the other similar function that does this

builds.sh
```diff
@@ -160,11 +183,20 @@ for sourceId; do
...
-			resolved="$(_resolve "$img" "$arch")"
+
+			if [ -n "${imageArchResolved[$img]:+x}" ]; then
+				remoteArches="${imageArchResolved["$img"]}"
+			else
+				remoteArches="$(_resolveRemoteArch "$img")"
+				imageArchResolved["$img"]="$remoteArches"
+			fi
+
+			resolved="$(_resolve "$img" "$arch" "$remoteArches")"
```
tianon, line 199:
```suggestion
			if [ -z "${imageArchResolved["$img"]:+x}" ]; then
				remoteArches="$(_resolveRemoteArch "$img")"
				imageArchResolved["$img"]="$remoteArches"
			fi

			resolved="$(_resolve "$img" "$arch" "${imageArchResolved["$img"]}")"
```

builds.sh
```diff
@@ -130,11 +146,18 @@ for sourceId; do
...
+				remoteArches="${imageArchResolved["$lookup"]}"
+			else
+				remoteArches="$(_resolveRemoteArch "$lookup")"
+				imageArchResolved["$lookup"]="$remoteArches"
+			fi
+
 			# if "$lookup" is a valid/known sourceId, we should look up the (pre-resolved) imageId for it
 			if [ -n "$lookup" ] && [ -n "${sources["$lookup"]:+x}" ]; then
 				resolved="${sourceArchResolved["$lookup-$arch"]:-}"
 			elif [ -n "$lookup" ]; then
```
tianon, line 159:
```suggestion
			# if "$lookup" is a valid/known sourceId, we should look up the (pre-resolved) imageId for it
			if [ -n "$lookup" ] && [ -n "${sources["$lookup"]:+x}" ]; then
				resolved="${sourceArchResolved["$lookup-$arch"]:-}"
			elif [ -n "$lookup" ]; then
				if [ -z "${imageArchResolved["$lookup"]:+x}" ]; then
					remoteArches="$(_resolveRemoteArch "$lookup")"
					imageArchResolved["$lookup"]="$remoteArches"
				fi
```

builds.sh
```diff
@@ -130,11 +146,18 @@ for sourceId; do
...
-				resolved="$(_resolve "$lookup" "$arch")"
+				resolved="$(_resolve "$lookup" "$arch" "$remoteArches")"
```
tianon, line 160:
```suggestion
				resolved="$(_resolve "$lookup" "$arch" "${imageArchResolved["$lookup"]}")"
```

tianon (commented):
Sorry! :see_no_evil:

Was giving this one more pass with the intent to merge, and realized we should move the first conditional down inside where the variable is used (so we avoid the remote lookup block entirely if we already have a cached value inside our `sources` variable), and took the opportunity to squeeze in a minor consistency change to both usages too. :sweat_smile:

LaurentGoderre:
OK, done. I appreciate the feedback btw!

tianon:
Did you forget to push? :sweat_smile:

LaurentGoderre:
@tianon  uuugh!! no.... when the repo went public it broke the link to my fork. I need to push directly to this repo now.

tianon (approved):

PR #5, opened 2023-12-07
Port builds.sh to Go (huge speed increase)

tianon:
I started by doing a naive port and got a very rough ~2x increase (tens of minutes down to just minutes), and then added a bit of parallelism (still roughly the same overall implementation -- if you want to review that commit, I'd suggest doing so with ignored whitespace) and managed to get down to seconds.  It does run into 429s periodically, which I think is probably expected (and could be mitigated by further caching the `oisupport/staging-XXX:YYY` tag lookups more aggressively, since they will only change if we're deleting them for an explicit cache bust / rebuild).  Additionally, this preserves the output order.

~~I've done a before/after on the generated file, and all the differences are what I'd expect -- the calculated buildIds for `drupal` and `notary` change, which will cause them to rebuild.  That happens because the old implementation was able to be order-preserving on parent -> digest mappings that get included in the buildId calculation, where Go doesn't have order-preserving maps so we get sorted parent keys instead (and thus a few new buildIds, but mostly 100% unchanged).  For builders that still preserve build cache (everything not on GHA), those should even be fully cached builds.~~
(added a simple "ordered map" implementation that avoids `buildId` changes)


tianon:
(https://github.com/golang/go/issues/27179 is a useful Go link)

tianon:
I'd like to get https://github.com/docker-library/meta-scripts/pull/7 in before this, then rebase this on that -- I'm still considering whether it's worth implementing ordered maps via a custom type so we can get back to the ideal state (I tried a few libraries but didn't really like any of them), but either way that'll help us make sure this is correct (while showing off how much faster it is via the job being faster) _and_ if we decide the `buildId` break is acceptable will show that in the diff (via updated "canonical output" in that new test).

tianon:
Writing a generic JSON round-tripping "ordered map" implementation that meets our needs was pretty straightforward (less than 100 lines total), so now we don't even have to live with a change to any of our `buildId`s :sweat_smile:

builds.go
```diff
@@ -0,0 +1,293 @@
...
+
+	"github.com/docker-library/meta-scripts/om"
+
+	c8derrdefs "github.com/containerd/containerd/errdefs"
+	"github.com/docker-library/bashbrew/registry"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus" // this is used by containerd libraries, so we need to set the default log level for it
+)
+
+var concurrency = 50
```
whalelines, line 22: Why `var` and not `const`?

builds.go
```diff
@@ -0,0 +1,293 @@
...
+	rArches := cache.arches
+
+	if _, ok := rArches[arch]; !ok {
+		// TODO this should probably be just like a 404, right? (it's effectively a 404, even if it's not literally a 404)
+		return nil, fmt.Errorf("%s missing %s arch", img, arch)
+	}
+	// TODO warn/error on multiple entries for arch? (would mean something like index/manifest list with multiple os.version values for Windows - we avoid this in DOI today, but we don't have any automated *checks* for it, so the current state is a little precarious)
+
+	ref := func(obj *registry.ResolvedObject) string {
+		base, _, _ := strings.Cut(obj.ImageRef, "@")
```
whalelines, line 111: Unassigned values necessary?

builds.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+
+			build.Build.SourceID = source.SourceID
+
+			if len(source.Arches) != 1 {
+				panic("unexpected arches length: " + string(build.Source))
+			}
+			for build.Build.Arch = range source.Arches {
+				// I really hate Go.
+				// (just doing a lookup of the only key in my map into a variable)
```
whalelines, line 186: Perhaps soon, https://pkg.go.dev/golang.org/x/exp/maps#Keys

builds.sh
```diff
@@ -1,228 +1,19 @@
...
-jq <<<"$builds" .
+dir="$(dirname "$BASH_SOURCE")"
+dir="$(readlink -ve "$dir")"
+if [ "$dir/builds.go" -nt "$dir/builds" ] || [ "$dir/om/om.go" -nt "$dir/builds" ] || [ "$dir/.go-env.sh" -nt "$dir/builds" ]; then
+	{
+		echo "building '$dir/builds' from 'builds.go'"
+		"$dir/.go-env.sh" go build -v -o builds builds.go
+		ls -l "$dir/builds"
+	} >&2
+fi
```
whalelines, line 16: `Makefile`?

om/om.go
```diff
@@ -0,0 +1,97 @@
...
+func (m OrderedMap[T]) MarshalJSON() ([]byte, error) {
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	if err := buf.WriteByte('{'); err != nil {
+		return nil, err
+	}
+	for i, key := range m.keys {
+		if i > 0 {
+			buf.WriteByte(',')
+		}
```
whalelines, line 84: Nice

om/om_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/docker-library/meta-scripts/om"
+)
+
+func assert[V comparable](t *testing.T, v V, expected V) {
+	t.Helper()
```
whalelines, line 11: I did not know about `Helper()`.

builds.go
```diff
@@ -0,0 +1,293 @@
...
+			fmt.Print(",")
+		} else {
+			first = false
+		}
+		fmt.Println()
+		buildId, err := json.Marshal(out.buildId)
+		if err != nil {
+			panic(err)
+		}
+		fmt.Printf("\t%s: %s", string(buildId), string(out.json))
```
whalelines, line 289: Why the tab character?

builds.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	fmt.Println()
```
whalelines, line 291: Why newline here? You do not appear to be be pretty printing otherwise.

whalelines (approved):
This is amazing, thanks!

Just a few minor nits and some questions.

builds.go
tianon: Just because I moved it up so it'd be more obvious and didn't give it much thought -- I ultimately wanted to make it a CLI flag, but changing it doesn't really change very much (it makes the whole thing slightly faster or slower, but still generally less than a minute so still subject to / limited by the 429s which are based on requests per minute). :sweat_smile:

builds.go
```diff
@@ -0,0 +1,293 @@
...
+	cacheResolve = sync.Map{}
+)
+
+type cacheResolveType struct {
+	r      *registry.ResolvedObject
+	arches map[string][]registry.ResolvedObject
+}
+
+func resolveRemoteArch(ctx context.Context, img string, arch string) (*RemoteResolvedFull, error) {
+	cacheFunc, _ := cacheResolve.LoadOrStore(img, sync.OnceValues(func() (*cacheResolveType, error) {
```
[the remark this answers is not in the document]
tianon:
```console
./builds.go:73:15: assignment mismatch: 1 variable but cacheResolve.LoadOrStore returns 2 values
```

builds.go
tianon:
```console
./builds.go:111:11: assignment mismatch: 1 variable but strings.Cut returns 3 values
```

builds.go
```diff
@@ -0,0 +1,293 @@
...
+
+	type out struct {
+		buildId string
+		json    []byte
+	}
+	outs := make(chan chan out, concurrency) // we want the end result to be "in order", so we have a channel of channels of outputs so each output can be generated async (and write to the "inner" channel) and the outer channel stays in the input order
+
+	go func() {
+		// Go does not have ordered maps *and* is complicated to read an object, make a tiny modification, write it back out (without modelling the entire schema), so we'll let a single invocation of jq solve both problems (munging the documents in the way we expect *and* giving us an in-order stream)
+		jq := exec.Command("jq", "-c", ".[] | (.arches | to_entries[]) as $arch | .arches = { ($arch.key): $arch.value }", sourcesJsonFile)
```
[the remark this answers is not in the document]
tianon: The order in `sources.json` is actually significant (it's what allows us to later assume that a parent referenced by `sourceId` is either already calculated or already in-progress so we can either wait for it or have it already), and if we sorted it we'd lose that property.  We could change the format of `sources.json` to be a list or something, but given that we _can_ have order-preserving and the cost of this is effectively nothing, it doesn't seem worth doing so IMO given that it being a map gives us the benefit of direct lookups by sourceId/buildId (which we then use in our builds).

That also only solves the ordering problem, not the "read, modify in a small way, write back out" problem (we'd have to define a _full_ schema for that sources object, which we'll need eventually if we port `sources.sh` into Go as well, but we don't need right now for the current problem).

builds.go
tianon: Indeed!  I was angry that https://pkg.go.dev/maps didn't include that (yet), but perhaps there is still hope.

builds.sh
tianon: I thought about it, but I'm not sure it actually makes this much simpler since it still has to list all the files that should cause the binary to rebuild, and it's really creating the container (via `.go-env.sh`) that I was trying to optimize not doing unless necessary, not the `go build` (which is otherwise _really_ fast and I've have no problem just running unconditionally every time).  I also don't want to encode an assumption of `make` being available on the host in scripts like this (as a general rule), so even with a `Makefile` I'd still be invoking it via `.go-env.sh` and be back to the same problem. :see_no_evil:

Instead of maintaining another file that then splits the logic for these things, I'd probably instead go the other direction and just pay the cost of the container overhead on every startup (since it should typically only be _real_ overhead the first time, when it has to pull the image).

builds.go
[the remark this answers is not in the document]
whalelines: gostaticcheck just told me to remove those in some situation. I guess that was in one of the places go allows that and this is not one of them. Why go, why?

builds.go
tianon: I'm going to answer this one and "Why the tab character?" together -- I _am_ roughly pretty-printing!  The value objects here are really large and deeply nested, so they are a _lot_ when they're pretty-printed to the terminal, but it _is_ useful to have the entries of the outer map be line-based, so this code has output that looks roughly like this:

```json
{
	"buildId-here": { ... large object here, on a single line ... },
	...
}
```

In other words, the load-bearing output thus resembles pretty closely the progress output (one line per build).

We pretty print the end result outside this for better diffs in Git (because Git's diffing algorithms generally do a better job with more newlines vs less).

whalelines: I forgot `out.json` had a newline at the end, so I thought everything ended up on a single line, with tabs between the key and the previous value. But now that I remember the JSON has a newline at the end, the `fmt.Println()` is _not_ necessary, right?

So the output ends up looking like this?

```json
{	"build-id-0": {…}
,	"build-id-1": {…}
…
,	"build-id-n": {…}

}
```

tianon: Nope - the `fmt.Printf` above does _not_ include a newline explicitly so that `,` can be on the line above (because that looks much cleaner / ergonomic):

```console
$ BASHBREW_STAGING_TEMPLATE=foo:BUILD ./builds <(printf '{}')
{
}
$ BASHBREW_STAGING_TEMPLATE=tianon/zz-staging:ARCH-BUILD ./builds <(BASHBREW_ARCH_NAMESPACES= BASHBREW_NAMESPACE=tianon BASHBREW_LIBRARY="$HOME/docker/tianon-bashbrew/library" ./sources.sh true 2>/dev/null) 2>/dev/null
{
	"93918980eb5f67dc11efc857eba1f0a33d9f433485d3d6b16101fb81cff85bb7": {"buildId":"93918980eb5f67dc11efc857eba1f0a33d9f433485d3d6b16101fb81cff85bb7","build":{"img":"tianon/zz-staging:amd64-93918980eb5f67dc11efc857eba1f0a33d9f433485d3d6b16101fb81cff85bb7","resolved":{"manifest":{"ref":"tianon/zz-staging:amd64-93918980eb5f67dc11efc857eba1f0a33d9f433485d3d6b16101fb81cff85bb7@sha256:0d18be811527e931c717c17bd130b37cd63cc15861502a1ee623cf6108aee783","desc":{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","digest":"sha256:0d18be811527e931c717c17bd130b37cd63cc15861502a1ee623cf6108aee783","size":524,"platform":{"architecture":"amd64","os":"linux"}}}},"sourceId":"f5f5c9e6e11fd02cff53245b41366e4e4b1c5a1c043a1d96ff4eccdb19194abc","arch":"amd64","parents":{},"resolvedParents":{}},"source":{"sourceId":"f5f5c9e6e11fd02cff53245b41366e4e4b1c5a1c043a1d96ff4eccdb19194abc","reproducibleGitChecksum":"dd443a9eeed44b2f73674af0954ae79ee4efd1b65dfbe35f5911914ec8b1bf18","allTags":["tianon/true:latest"],"entry":{"GitRepo":"https://github.com/tianon/dockerfiles.git","GitFetch":"refs/heads/master","GitCommit":"bfa71889cd8e48e7a1d2632422785c479bd9bd45","Directory":"true","File":"Dockerfile","Builder":"","SOURCE_DATE_EPOCH":1675234271},"arches":{"amd64":{"tags":["tianon/true:latest"],"archTags":[],"froms":["scratch"],"platformString":"linux/amd64","platform":{"architecture":"amd64","os":"linux"},"parents":{"scratch":{"sourceId":null,"pin":null}}}}}},
	"ea8772845de98e529b90034e19eab1979ebbe697963cbafa97c823d06eb32fa8": {"buildId":"ea8772845de98e529b90034e19eab1979ebbe697963cbafa97c823d06eb32fa8","build":{"img":"tianon/zz-staging:amd64-ea8772845de98e529b90034e19eab1979ebbe697963cbafa97c823d06eb32fa8","resolved":{"manifest":{"ref":"tianon/zz-staging:amd64-ea8772845de98e529b90034e19eab1979ebbe697963cbafa97c823d06eb32fa8@sha256:125a236c3f02370d5110e7dce47664d99f4f4a3da00151f0bf9fb8568be4e398","desc":{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","digest":"sha256:125a236c3f02370d5110e7dce47664d99f4f4a3da00151f0bf9fb8568be4e398","size":524,"platform":{"architecture":"amd64","os":"linux"}}}},"sourceId":"ffc37b0864706846f1f9f014df57ac0572ab9dd9eb99f5f0d318b26cd19f4721","arch":"amd64","parents":{},"resolvedParents":{}},"source":{"sourceId":"ffc37b0864706846f1f9f014df57ac0572ab9dd9eb99f5f0d318b26cd19f4721","reproducibleGitChecksum":"203c5d023e44b5df1f46d41df253a770ed508c672738625b91ff3f464193fbdb","allTags":["tianon/true:yolo"],"entry":{"GitRepo":"https://github.com/tianon/dockerfiles.git","GitFetch":"refs/heads/master","GitCommit":"47fe0d0cb5ede5a36907162c25b058fd926f22fc","Directory":"true","File":"Dockerfile.yolo","Builder":"","SOURCE_DATE_EPOCH":1677043314},"arches":{"amd64":{"tags":["tianon/true:yolo"],"archTags":[],"froms":["scratch"],"platformString":"linux/amd64","platform":{"architecture":"amd64","os":"linux"},"parents":{"scratch":{"sourceId":null,"pin":null}}}}}}
}
```

whalelines: My internal compiler toolchain is a bit off. I missed https://github.com/docker-library/meta-scripts/pull/5/files#diff-293e1f9ac7fe0e3a8531eedbc7cd2b8772bc0dda2ae9497f288bfa1a6cc4b3f4R284 , and it looks like https://github.com/docker-library/meta-scripts/pull/5/files#diff-293e1f9ac7fe0e3a8531eedbc7cd2b8772bc0dda2ae9497f288bfa1a6cc4b3f4R232 does not end up the this JSON.

whalelines (approved):
LGTM, sorry for the hassle

builds.go
tianon: Ahh, those functions being included in stdlib is blocked on the "iterators" proposals (which are blocked on "`for range` over a function" that was accepted but still experimental and requires `GODEBUG` incantations to use), but an iterator would honestly be even better for this use case than a slice of keys so I guess I'll sit down and try to be patient. :joy:

yosifkit (approved):

PR #6, opened 2023-12-08
Add initial GitHub Actions tests of behavior/output

tianon:


tianon:
Oh right, first PR of a new action has to be on the origin.  Thanks GitHub! :upside_down_face:

PR #4, opened 2023-11-15
Add ability to generate builds.json in parallel

LaurentGoderre:

PR #7, opened 2023-12-08
Add initial GitHub Actions tests of behavior/output

tianon:
(https://github.com/docker-library/meta-scripts/pull/6, take 2)


tianon:
To unpack what I've done here a little:

- copied `library/notary` and `library/docker` verbatim (so they're pinned / reproducible)
- copied the couple `.external-pins/*.sh` scripts we need to accomplish image pinning
- explicitly pinned all parents of `notary`, `docker:cli`, `docker:dind`, and `docker:windowsservercore` (except `golang:1.19-alpine3.16`, which is EOL and thus should not be updated and is a perfect test of the "parent neither pinned nor in our set of things to build" edge case)

All these have valid builds, so even the `oistaging` lookups succeed correctly and as expected.

Thus, we're able to have "canonical" versions of `sources.json` and `builds.json` which are fully reproducible that get committed so they can be compared against in a re-run.

I haven't tested it directly, but this _should_ be enough to even catch what I was describing over in #4 :eyes:

yosifkit (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: 0 0 * * 0
```
whalelines, line 7: Why only on Sunday?

tianon: I put something like this in most of my GHA scripts as a fallback "in case it hasn't run this week, do a run on Sunday when nobody's likely to be looking at it to make sure it's still green" because almost everything I write ends up having to be integration tests (and thus prone to breaking due to external factors that don't necessarily create a commit in the repository).

If GHA had a way to say "run this at a random time Soon if it hasn't run in 14 days" or something like that, I'd happily switch to it. :sweat_smile:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: 'bash -Eeuo pipefail -x {0}'
```
whalelines, line 15: What does the `{0}` do here?

.test/.external-pins/library/alpine___3.16
```diff
@@ -0,0 +1 @@
+sha256:e4cdb7d47b06ba0a062ad2a97a7d154967c8f83934594d9f2bd3efa89292996b
```
whalelines, line 1: Aren't these versions out of support?

tianon: Alpine 3.16 is still supported (it's 3.15 that just went EOL), but pinning like this ensures our tests are reproducible (we'll be treating `alpine:3.16` as `sha256:e4cdb7d47b06ba0a062ad2a97a7d154967c8f83934594d9f2bd3efa89292996b` in the future for the purposes of these tests even if it's updated).

PR #8, opened 2023-12-13
Update "meta" job to skip "Sources" if `sources.json` is already up-to-date

tianon:
The contents of `sources.json` should be really static unless something in the scripts or the upstream sources has changed, so we only need to pay that multi-minute cost in cases where the scripts or source data has actually changed.

There are possibly edge cases here where these upstream things have changed and `sources.json` hasn't/won't. :thinking:

However, this _is_ strictly better than the current state where we just run it all the time.

I think we could probably avoid some of the other edge cases by having this whole script _only_ commit updates to our `.doi` and `.scripts` submodules if `sources.json` or `builds.json` actually have changes. :thinking:

### DIFF
--- a/Jenkinsfile.update
+++ b/Jenkinsfile.update
@@ -48,7 +48,27 @@ node {
 				sh 'bashbrew --library .doi/library fetch $(cat subset.txt)' // TODO --all
 			}
 			stage('Sources') {
-				sh '.scripts/sources.sh $(cat subset.txt) > sources.json'
+				sh '''
+					# we only need to regenerate "sources.json" if ".doi", ".scripts", or "subset.txt" have changed since we last generated it
+					needsUpdate=
+					if ! git diff --exit-code .doi .scripts; then
+						# if .doi or .scripts updated in this job ("git submodule update" above)
+						needsUpdate=1
+					else
+						# if subset.txt, .doi, or .scripts updated outside this job (merged PR, etc)
+						sourcesCommit="$(git log -1 --format='format:%H' -- sources.json)"
+						for f in subset.txt .doi .scripts; do
+							commit="$(git log -1 --format='format:%H' -- "$f")"
+							if ! git merge-base --is-ancestor "$commit" "$sourcesCommit"; then
+								needsUpdate=1
+								break
+							fi
+						done
+					fi
+					if [ -n "$needsUpdate" ]; then
+						.scripts/sources.sh $(cat subset.txt) > sources.json
+					fi
+				'''
 			}
 			stage('Builds') {
 				sh '.scripts/builds.sh sources.json > builds.json'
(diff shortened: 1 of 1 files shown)


tianon:
For reference, merging https://github.com/docker-library/meta/pull/10 took our "Sources" stage from ~2m30s up to ~3m20s and this change would bring it back down to milliseconds for any run that can prove that it doesn't have to actually update `sources.json`.

tianon:
> I think we could probably avoid some of the other edge cases by having this whole script _only_ commit updates to our `.doi` and `.scripts` submodules if `sources.json` or `builds.json` actually have changes. 🤔

Ok, this is implemented now too.

Jenkinsfile.update
```diff
@@ -48,16 +48,39 @@ node {
...
 			}
 		}
 		stage('Commit') {
 			sh '''
-				git add -A .
-				git commit -m 'Update and regenerate' || :
+				# only commit submodule updates if our JSON has changed (see https://github.com/docker-library/meta-scripts/pull/8)
+				if ! git diff --exit-code sources.json builds.json; then
+					git add -A .
+					git commit -m 'Update and regenerate' || :
```
whalelines, line 82: Why is it okay if the commit fails?

whalelines (approved):
LGTM, just one question.

Jenkinsfile.update
tianon: `git commit` fails if there's nothing to commit, which is fair (and `git push` will automatically be a no-op in that case)

Now that I'm gating this on JSON files having changes, we should probably let `git commit` failing fail the build instead, so that's a good call-out :+1:

yosifkit (approved):

tianon:
There's another edge case here that means we no-op regenerated this file all night long.  If there are changes to `.doi` but they do _not_ result in changes to `sources.json`, we will re-generate on every subsequent run until that changes.

I thought about this one while writing it, and my conclusion was that it's probably OK to ignore this for now because it will improve over time as we add more images and it becomes more likely that changes to `.doi` actually do result in changes to `sources.json`, and this was intended as an optimization anyhow so falling back to the older behavior is pretty reasonable.

That being said, we could try harder here to resolve this, if we wanted to.

We could get more specific about _which_ things changing in `.doi` will trigger this, with the major downside that it's a game of whack-a-mole that might result in us not regenerating when we should because something is relevant and we forgot or missed it for some reason.  IMO, this needs to always fail on the side of regenerating more often than necessary vs _missing_ generation runs (because the former is "just" a CPU cost and the latter leads to missing builds / missing images / confused and annoyed users).

A simpler option would be record in another file which commit of `.doi` and `.scripts` generated the committed `sources.json` so that if there's a new commit that results in no changes to `sources.json`, we still _try_ generating it at least once, but then record that we did so such that we do not have to do so again.  (In case the follow-up question to this is why not just use the submodule commits as this reference, I'm worried that updating submodule commits is something that's too easy to do accidentally, so I'd rather it be something explicitly generated and committed so it's a more clear signal of intent / state.)

While typing this, that compromise seems pretty reasonable, so I'll probably work on that later today. :sweat_smile:

PR #9, opened 2023-12-14
Implement `commits.json` that tracks whether `sources.json` is potentially outdated

tianon:
https://github.com/docker-library/meta-scripts/pull/8#issuecomment-1856298144

> A simpler option would be record in another file which commit of `.doi` and `.scripts` generated the committed `sources.json` so that if there's a new commit that results in no changes to `sources.json`, we still _try_ generating it at least once, but then record that we did so such that we do not have to do so again.  (In case the follow-up question to this is why not just use the submodule commits as this reference, I'm worried that updating submodule commits is something that's too easy to do accidentally, so I'd rather it be something explicitly generated and committed so it's a more clear signal of intent / state.)

### DIFF
--- a/Jenkinsfile.update
+++ b/Jenkinsfile.update
@@ -50,23 +50,25 @@ node {
 			stage('Sources') {
 				sh '''
 					# we only need to regenerate "sources.json" if ".doi", ".scripts", or "subset.txt" have changed since we last generated it
-					needsUpdate=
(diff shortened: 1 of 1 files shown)


whalelines (approved):
LGTM

yosifkit (approved):

tianon:
Confirmed working :muscle:

PR #10, opened 2023-12-15
Retry each crane command on 429 after a sleep

yosifkit:
Also, move the crane commands generator to a `.jq` file instead of embedded in the Jenkinsfile for easier debugging

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -64,21 +64,8 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 				set -Eeuo pipefail -x
 
 				jq -L.scripts -r '
-					include "meta";
-					reduce (.[] | select(.build.resolved and .build.arch == env.BASHBREW_ARCH)) as $i ({};
-						.[ $i.source.arches[].archTags[] ] += [
-							$i.build.resolved
-							| .index.ref // .manifest.ref
-						]
-					)
-					| to_entries[]
-					| .key as $target
-					| .value
-					| if length == 1 then
-						@sh "crane copy \\(.) \\($target)"
-					else
-						@sh "crane index append --tag \\($target) " + (map("--manifest " + @sh) | join(" ")) + " --flatten"
-					end
+					include "jenkins";
+					crane_deploy_commands | @sh
 				' builds.json
 			''').trim()
 
@@ -89,7 +76,24 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			sh """#!/usr/bin/env bash
 				set -Eeuo pipefail
 
-				${ shell }
+				commands=( ${ shell } )
+				for c in "${commands[@]}"; do
+					tries=3
+					while ! output="$( { $c; } |& tee /dev/stderr )"; then
+						# check to see if we hit the docker hub rate limit
+						if grep -q 'TOOMANYREQUESTS' <<<"$output"; then
+							if [ "$(( --tries ))" -le 0 ]; then
+								echo >&2 'failed to copy image 3 times'
+								exit 1
+							fi
+							echo 'sleeping before retrying...'
+							sleep 60
+						else
+							# non 429 error
+							exit 1
+						fi
+					fi
+				done
 			"""
 		}
 	}
--- a/jenkins.jq
+++ b/jenkins.jq
@@ -1,0 +1,18 @@
+# input: list of build objects i.e., builds.json
+# output: stream of crane copy command strings
+def crane_deploy_commands:
+	reduce (.[] | select(.build.resolved and .build.arch == env.BASHBREW_ARCH)) as $i ({};
+		.[ $i.source.arches[].archTags[] ] += [
+			$i.build.resolved
+			| .index.ref // .manifest.ref
+		]
+	)
+	| to_entries[]
+	| .key as $target
+	| .value
+	| if length == 1 then
+		@sh "crane copy \(.) \($target)"
+	else
+		@sh "crane index append --tag \($target) " + (map("--manifest " + @sh) | join(" ")) + " --flatten"
+	end
+;


Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			sh """#!/usr/bin/env bash
 				set -Eeuo pipefail
 
-				${ shell }
+				commands=( ${ shell } )
+				for c in "${commands[@]}; do
```
whalelines, line 80:
```suggestion
				for c in "${commands[@]}"; do
```

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			sh """#!/usr/bin/env bash
 				set -Eeuo pipefail
 
-				${ shell }
+				commands=( ${ shell } )
```
whalelines, line 79: Won't this split the expanded Groovy variable `shell` on whitespace?

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
...
+					tries=3
+					while ! output="$( { $c; } |& tee /dev/stderr )"; then
+						# check to see if we hit the docker hub rate limit
+						if grep -q 'TOOMANYREQUESTS' <<<"$output"; then
+							(( tries-- ))
+							if [ "$iter" -le 0 ]; then
+								echo >&2 'failed to copy image 3 times'
+								exit 1
+							fi
+							echo 'sleeping...'
```
whalelines, line 90:
```suggestion
							echo 'sleeping before retrying...'
```

whalelines (commented):
This is a great idea. Just a couple questions…

Jenkinsfile.deploy
yosifkit: Doh, yes.

Jenkinsfile.deploy
yosifkit: Yup. That's why it does `crane_deploy_commands | @sh` to create it. The function will return a stream of crane commands and then each one will be shell quoted to be a single shell string.

```
crane_deploy_commands
crane copy ...
crane copy ...
-> @sh
'crane copy ...'
'crane copy ...'
```

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
...
 
-				${ shell }
+				commands=( ${ shell } )
+				for c in "${commands[@]}"; do
+					tries=3
+					while ! output="$( { $c; } |& tee /dev/stderr )"; then
+						# check to see if we hit the docker hub rate limit
+						if grep -q 'TOOMANYREQUESTS' <<<"$output"; then
+							(( tries-- ))
+							if [ "$iter" -le 0 ]; then
```
whalelines, line 86:
```suggestion
							if [ "$tries" -le 0 ]; then
```

whalelines (approved):
LGTM

Jenkinsfile.deploy
tianon: And to be clear, this won't be shell doing expansion/splitting, but rather the Groovy will embed the generated and appropriately quoted shell code directly in the string it passes to the shell for invocation, so the end result of what the shell actually executes will be literally something like:

```bash
commands=( 'crane copy ...'
'crane copy ...' )
```

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
...
 				set -Eeuo pipefail
 
-				${ shell }
+				commands=( ${ shell } )
+				for c in "${commands[@]}"; do
+					tries=3
+					while ! output="$( { $c; } |& tee /dev/stderr )"; then
+						# check to see if we hit the docker hub rate limit
+						if grep -q 'TOOMANYREQUESTS' <<<"$output"; then
+							(( tries-- ))
```
tianon, line 85:
```suggestion
							(( tries-- )) || :
```

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
...
+							if [ "$tries" -le 0 ]; then
```
tianon, line 86: Alternatively:

```suggestion
							if [ "$(( --tries ))" -le 0 ]; then
```

jenkins.jq
```diff
@@ -0,0 +1,18 @@
+# input: list of build objects i.e., builds.json
+# output: stream of crane copy command strings
+def crane_deploy_commands:
+	reduce (.[] | select(.build.resolved and .build.arch == env.BASHBREW_ARCH)) as $i ({};
```
tianon, line 4: In the future, it might be worth splitting this select out to be the responsibility of the caller, but I don't think we should do that here :eyes:

(Especially since the use cases I have in mind aren't well formed yet, so it's not worth doing anything concrete yet)

Jenkinsfile.deploy
```diff
@@ -89,7 +76,25 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			sh """#!/usr/bin/env bash
 				set -Eeuo pipefail
 
-				${ shell }
+				commands=( ${ shell } )
+				for c in "${commands[@]}"; do
+					tries=3
+					while ! output="$( { $c; } |& tee /dev/stderr )"; then
+						# check to see if we hit the docker hub rate limit
+						if grep -q 'TOOMANYREQUESTS' <<<"$output"; then
```
tianon, line 84: Is `TOOMANYREQUESTS` something `crane` outputs, or something that just happens to be in the error text from Docker Hub right now (and thus might change in the future)?

I don't think this is worth doing much about (ie, this is fine as-is because the failure mode would be that we just ... don't retry and thus the problem becomes obvious), but something to consider in the future.

tianon (commented):
Just one minor nit and a few non-action comments :+1:

tianon:
Fire in the hole!

tianon:
We're seeing some weirdness in how this actually runs, and it's probably thanks to https://www.docker.com/increase-rate-limits/ (these jobs are actually hitting the "normal" 429 that only resets every 6 hours and for the user on that host only gives 200 / 6 hours and our other usage on that system has been mitigating that in other ways until now :upside_down_face:).

Gonna look at patching our `crane` build now to see if we can have an "easy" bandaid.

tianon:
With a bit more hacking, this is now doing much better (the bandaid was extremely successful). :+1:

PR #13, opened 2024-01-11
Implement optional persistent registry lookup cache and better 429 handling

tianon:
This adds a `--cache` flag to builds which will read from a cache file and write back out to it a set of images that are ~safe to cache (and because it's just a file, we can clear cache entries trivially by deleting the relevant entries from the file).

This also adds better 429 behavior that's twofold:

1. cap our maximum registry requests at ~500/min with an initial burst of 100 unless we hit a 429 (which should, in theory, keep us closer to staying under the Hub abuse rate limits)

2. actually retry requests when they give us a 429 response (capping both new requests and retries at the same ~500/min which is ~8/sec)

As a final bonus, I added/generated an appropriate cache file for our local test suite which brings the total number of actual Hub/registry requests our test suite makes down to one single image lookup.


yosifkit (approved):

builds.go
```diff
@@ -4,22 +4,26 @@ import (
...
 	c8derrdefs "github.com/containerd/containerd/errdefs"
+	c8derrs "github.com/containerd/containerd/remotes/errors"
 	"github.com/docker-library/bashbrew/registry"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus" // this is used by containerd libraries, so we need to set the default log level for it
+	"golang.org/x/time/rate"
 )
 
-var concurrency = 50
+var concurrency = 1000
```
whalelines, line 26: How does this interact with the rate limiter?

builds.go
```diff
@@ -128,9 +198,97 @@ func resolveRemoteArch(ctx context.Context, img string, arch string) (*RemoteRes
...
+	}
+
+	return nil
+}
+
+func saveCacheToFile() error {
+	saveCacheMutex.Lock()
+	defer saveCacheMutex.Unlock()
+
+	if saveCache == nil || cacheFile == "" {
```
whalelines, line 256: Could this be moved before locking the mutex?

builds.go
tianon: This was implemented initially as a poor-man's rate limiter, so now that we have a "real" rate limiter in place, we could technically remove this completely and let it be completely unbounded, but I figured making it high (instead of unlimited) was probably better/safer.

In practice the rate limiter is likely to block many of these goroutines anyhow (and having them wait to make a request vs wait to start seems totally fair since the overhead is really low and they can potentially do other things in the mean time).  Also, with the cache implemented, many of the goroutines won't block at all, so it's just "free" concurrency. :+1:

builds.go
tianon: The `cacheFile` test could, but the test for `saveCache` cannot because we cannot safely compare that value without a mutex (even a simple comparison like `nil`).

(This is also effectively the test for "did the user specify `--cache`", and we're specifying `--cache` in all the places we care about now, so this test is effectively as good as a no-op now.)

builds.go
```diff
@@ -98,8 +151,25 @@ func resolveRemoteArch(ctx context.Context, img string, arch string) (*RemoteRes
...
 		return nil, nil
 	}
-	r := cache.r
-	rArches := cache.arches
+
+	r := cache.Resolved
+	rArches := cache.Arches
+
+	if !wasCached {
+		fmt.Fprintf(os.Stderr, "NOTE: lookup %s -> %s\n", img, r.Desc.Digest)
```
whalelines, line 159: Is this a debugging statement or indented to remain?

tianon: It's a bit of both -- I added it for debugging, but realized it was useful enough to know when lookups _actually_ happen that I left it in. :sweat_smile:

PR #14, opened 2024-01-17
Adjust rate limit to 100/min

tianon:
Also add a new limiter for individual lookups to top out retries at ~1/sec

This makes our absolute worst case really bad (like 70 minutes or so), but then the cache will make the follow-ups fast again.


LaurentGoderre (approved):

whalelines (approved):
LGTM

yosifkit (approved):

PR #15, opened 2024-01-18
Workaround "docker build giturl" requiring the url to end in ".git"

yosifkit:
Requirements like in [moby/buildkit `git_ref.go`](https://github.com/moby/buildkit/blob/0e1e36ba9eb8142968b2c5cfa2f12549bf9246d9/util/gitutil/git_ref.go#L81-L87) (even though git URLs don't need to end in `.git`).


meta.jq
```diff
@@ -53,10 +53,24 @@ def pull_command:
 	end
 ;
 # input: "build" object (with "buildId" top level key)
+# output: string "giturl" ("https://github.com/docker-library/golang#commit:directory), used for "docker build giturl"
```
tianon, line 56:
```suggestion
# output: string "giturl" ("https://github.com/docker-library/golang.git#commit:directory), used for "docker buildx build giturl"
```

meta.jq
```diff
@@ -53,10 +53,24 @@ def pull_command:
...
+def git_build_url:
+	.source.entry.GitRepo as $gitRepo
+	| if ($gitRepo | test("^https?://github.com/"))
+		and ($gitRepo | endswith(".git") | not) then
+			# without ".git" in the url "docker build url" fails and tries to build the html repo page as a Dockerfile
+			# https://github.com/moby/moby/blob/4f9c865eddc09da0c0cb9fe08c76b81b804093f5/builder/remotecontext/urlutil/urlutil.go#L15
```
tianon, line 62: (included in the better later suggestion)

```suggestion
			# without ".git" in the url "docker buildx build url" fails and tries to build the html repo page as a Dockerfile
			# https://github.com/moby/buildkit/blob/0e1e36ba9eb8142968b2c5cfa2f12549bf9246d9/util/gitutil/git_ref.go#L81-L87
```

meta.jq
```diff
@@ -53,10 +53,24 @@ def pull_command:
...
+			$gitRepo + ".git"
+	else
+		$gitRepo
+	end
```
tianon, line 66: whitespace nit (superseded by the better later suggestion)

```suggestion
		else
			$gitRepo
		end
```

meta.jq
```diff
@@ -53,10 +53,24 @@ def pull_command:
...
+	+ "#" + .source.entry.GitCommit + ":" + .source.entry.Directory
```
tianon, line 67: Even better we can simplify the whole block with less `()` by putting some on the outside and ditching the variable:

```suggestion
	.source.entry
	| (
		.GitRepo
		| if (endswith(".git") | not) then
			if test("^https?://github.com/") then
				# without ".git" in the url "docker buildx build url" fails and tries to build the html repo page as a Dockerfile
				# https://github.com/moby/buildkit/blob/0e1e36ba9eb8142968b2c5cfa2f12549bf9246d9/util/gitutil/git_ref.go#L81-L87
				. + ".git"
			else
				error("\(.) does not end in '.git' so build will fail to recognize it as a Git URL")
			end
		else . end
	) + "#" + .GitCommit + ":" + .Directory
```

(this also adds an error for anything we know `docker buildx build` will fail inscrutably with)

```console
$ jq -nc '"https://github.com/foo/bar" | if (endswith(".git") | not) then
			if test("^https?://github.com/") then
				# without ".git" in the url "docker buildx build url" fails and tries to build the html repo page as a Dockerfile
				# https://github.com/moby/buildkit/blob/0e1e36ba9eb8142968b2c5cfa2f12549bf9246d9/util/gitutil/git_ref.go#L81-L87
				. + ".git"
			else
				error("\(.) does not end in '.git' so build will fail to recognize it as a Git URL")
			end
		else . end'
"https://github.com/foo/bar.git"

$ jq -nc '"https://tianon.xyz/foo/bar" | if (endswith(".git") | not) then
			if test("^https?://github.com/") then
				# without ".git" in the url "docker buildx build url" fails and tries to build the html repo page as a Dockerfile
				# https://github.com/moby/buildkit/blob/0e1e36ba9eb8142968b2c5cfa2f12549bf9246d9/util/gitutil/git_ref.go#L81-L87
				. + ".git"
			else
				error("\(.) does not end in '.git' so build will fail to recognize it as a Git URL")
			end
		else . end'
jq: error (at <unknown>): https://tianon.xyz/foo/bar does not end in .git so build will fail to recognize it as a Git URL
```

tianon (changes requested):

tianon:
Aha! https://github.com/docker/cli/issues/1738

LaurentGoderre:
https://github.com/docker-library/official-images/pull/16082

meta.jq
whalelines: That regular expression is `".git(?:#.+)?$"`, which allows for anchors while `($gitRepo | endswith(".git") | not)` does not. Is this okay?

tianon (approved):

tianon:
Yay https://github.com/docker-library/meta/actions/runs/7574232353/job/20628160728 :tada:

meta.jq
whalelines: It appears that BuildKit code parses the URL and matches specifically on the Path portion of the URL, which does not include any query or fragment portions of the URL.

tianon: Yeah, I don't think `git clone` on a URL with a query parameter would actually work correctly, but maybe there's some esoteric server somewhere?  I'm not worried about the regex being `.git` because that's just a bug (a 6 year old bug, to be fair, so could be considered "load bearing", but BuildKit implemented the same check more correctly and nobody's noticed, so I don't think anyone's put real load on it)

tianon: (In other words, any `GitRepo:` with an anchor in it would be invalid already)

PR #16, opened 2024-01-20
Update sources.sh to use bashbrew 0.1.10's `ArchGitChecksum`

tianon:
This makes a *very* dramatic difference in the speed of `sources.sh`.

Before:

```console
$ time .scripts/sources.sh $(cat subset.txt) > sources.json
...
real	6m13.444s
user	2m51.334s
sys	3m29.256s
```

After:

```console
$ time .scripts/sources.sh $(cat subset.txt) > sources.json
...
real	0m31.238s
user	0m32.769s
sys	0m1.375s
```


tianon:
Gahhhhh, I forgot I also needed the `sha256sum` function :sob:

(That's not actually what I would've chosen to name it, but that's what [Sprig](https://github.com/docker-library/bashbrew/issues/22) names it, so in the spirit of avoiding further deviance from them I chose to follow suit.)

tianon:
https://github.com/docker-library/bashbrew/pull/90

tianon:
Alright, this should be _properly_ good to go now.

tianon:
With this it's probably ~safe to revert #9 / #8, but it feels better to take small steps (and that code is definitely working correctly ATM).

yosifkit (approved):

PR #17, opened 2024-01-25
Refactor `meta.jq` to prepare for signing and `oci-import` implementation

tianon:
- explicit "containerd image storage in dockerd" placeholder
- explicit "annotations" helper/generator (so they can be added in the `oci-import` builder later)
- setting the `org.opencontainers.image.created` annoation inside the image index to the actual build date (otherwise `SOURCE_DATE_EPOCH` makes it hard to tell an image is actually freshly built)
- more whitespace in generated commands for better readability
- extract OCI tar in the build step (so the "output" of the build is an OCI layout directory vs tarball) instead of the pull step (so we can more ergonomically add pre-push signing)
- use buildx's `--annotation` flag for making annotations easier to read

PR #18, opened 2024-01-25
Add "example" generated commands (so changes to generated commands are easier to see)

tianon:
See https://github.com/docker-library/meta-scripts/pull/17#issuecomment-1911144205

> Realizing we could (ab)use our `sources.json` and `builds.json` test suite to generate some example commands in the repo itself so that changes like this actually _show_ what the pull/build/push command change is (and helps illustrate why the changes are being made). :eyes:

### DIFF
--- a/.test/example-commands.sh
+++ b/.test/example-commands.sh
@@ -1,0 +1,27 @@
+# docker:24.0.7-cli [amd64]
+# <pull>
+
+# </pull>
+# <build>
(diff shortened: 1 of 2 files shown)


.test/example-commands.sh
```diff
@@ -0,0 +1,27 @@
+# docker:24.0.7-cli [amd64]
+# <pull>
+
+# </pull>
```
tianon, line 4: I could work harder to collapse this, but :shrug: (I didn't think it was worth it -- happy to adjust if you think it's worth doing, @yosifkit).

.test/test.sh
```diff
@@ -12,5 +12,27 @@ set -- docker:cli docker:dind docker:windowsservercore notary # a little bit of
 # (see "library/" and ".external-pins/" for where these come from / are hard-coded for consistent testing purposes)
 # NOTE: we are explicitly *not* pinning "golang:1.19-alpine3.16" so that this also tests unpinned parent behavior (that image is deprecated so should stay unchanging)
 
+time bashbrew fetch "$@"
```
tianon, line 15: Splitting this out from being implied by `sources.sh` would've made PRs like #16 more interesting (because we'd see the actual change in generation time, not affected by Git fetch time), but we're doing such a small set here that it's really not much of a difference. :smile:

.test/example-commands.sh
yosifkit: It seems fine.

yosifkit (approved):

.test/test.sh
```diff
@@ -12,5 +12,27 @@ set -- docker:cli docker:dind docker:windowsservercore notary # a little bit of
...
+
 time "$dir/../builds.sh" --cache "$dir/cache-builds.json" "$dir/sources.json" > "$dir/builds.json"
+
+# generate an "example commands" file so that changes to generated commands are easier to review
+jq -r -L "$dir/.." '
+	include "meta";
+	[
+		first(.[] | select(normalized_builder == "buildkit")),
+		first(.[] | select(normalized_builder == "classic")),
+		first(.[] | select(normalized_builder == "oci-import")),
```
whalelines, line 27: It seems like this would select three builds, but there are only two sets of commands generated. Is this because there are no oci-import examples yet?

tianon: Yes, that's correct -- I had originally left a `# TODO oci-import` here, but realized I could just add the conditional and it would match as soon as there was an example in the test data (which I plan to add when I have an implementation in `meta.jq`).

tianon: (In other words, this is currently a very fancy and computational way to spell `empty` :joy:)

PR #19, opened 2024-01-26
Set `org.opencontainers.image.base.name` (and `org.opencontainers.image.base.digest`) appropriately

tianon:
See https://github.com/opencontainers/image-spec/blob/v1.1.0-rc6/annotations.md#pre-defined-annotation-keys

See also https://github.com/opencontainers/image-spec/issues/821 and https://github.com/opencontainers/image-spec/pull/822

I noticed recently again that I'd implemented `ArchLastStageFrom` (which is what's necessary to set these correctly), _and_ thanks to #18 and #17, this implementation's correctness is easier to have confidence in. :smile:

### DIFF
--- a/.test/builds.json
+++ b/.test/builds.json
@@ -93,6 +93,7 @@
           "froms": [
             "alpine:3.18"
           ],
+          "lastStageFrom": "alpine:3.18",
           "platformString": "linux/amd64",
           "platform": {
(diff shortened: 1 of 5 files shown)


whalelines (approved):
LGTM

yosifkit (approved):

PR #20, opened 2024-01-30
Add initial `Builder: oci-import` support

tianon:
See also:
- https://github.com/docker-library/bashbrew/pull/51
- https://github.com/docker-library/bashbrew/pull/61
- https://github.com/docker-library/official-images/pull/13769

This includes a workaround for the bug it helped me find (and fix in https://github.com/docker-library/bashbrew/pull/92), but I don't think it needs to wait for the fix because the workaround is really trivial / straightforward.

The biggest benefit here (besides being one step closer to being able to switch to the new method for _everything_) is that this new way we build images will be significantly less susceptible to the downsides of the previous `oci-import` support (most notably, that the tarballs round-tripping through Docker recompresses them -- the containerd image storage in Docker itself should also help make that much better).

That also means this would make me feel comfortable with finally replying to https://github.com/docker-library/official-images/issues/527 to point them at the `oci-import` support that they might be interested in (and thus resolving https://github.com/docker-library/meta/pull/22 one way or another). :eyes:

### DIFF
--- a/.test/builds.json
+++ b/.test/builds.json
@@ -2716,5 +2716,61 @@
         }
       }
     }
+  },
+  "93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf": {
+    "buildId": "93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf",
+    "build": {
+      "img": "oisupport/staging-amd64:93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf",
+      "resolved": null,
+      "sourceId": "a7e4b56dd1e5dde4c9d988092cb8639987559f13c5c92210664a4ffb880f222b",
+      "arch": "amd64",
+      "parents": {},
+      "resolvedParents": {}
+    },
+    "source": {
+      "sourceId": "a7e4b56dd1e5dde4c9d988092cb8639987559f13c5c92210664a4ffb880f222b",
+      "reproducibleGitChecksum": "0ea17ff707666270e1fcb3edd76545906c68714d658ccde88c22d1606b2d7e4f",
+      "allTags": [
+        "ubuntu:22.04",
+        "ubuntu:jammy-20240111",
+        "ubuntu:jammy",
+        "ubuntu:latest"
+      ],
+      "entry": {
+        "GitRepo": "https://git.launchpad.net/cloud-images/+oci/ubuntu-base",
+        "GitFetch": "refs/tags/dist-jammy-amd64-20240111-e6e3490a",
+        "GitCommit": "e6e3490ad3f524ccaa072edafe525f8ca8ac5490",
+        "Directory": "oci",
+        "File": "index.json",
+        "Builder": "oci-import",
+        "SOURCE_DATE_EPOCH": 1704931200
+      },
+      "arches": {
+        "amd64": {
+          "tags": [
+            "ubuntu:22.04",
+            "ubuntu:jammy-20240111",
+            "ubuntu:jammy",
+            "ubuntu:latest"
+          ],
+          "archTags": [],
+          "froms": [
+            "scratch"
+          ],
+          "lastStageFrom": "scratch",
+          "platformString": "linux/amd64",
+          "platform": {
+            "architecture": "amd64",
+            "os": "linux"
+          },
+          "parents": {
+            "scratch": {
+              "sourceId": null,
+              "pin": null
+            }
+          }
+        }
+      }
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


meta.jq
```diff
@@ -265,10 +265,76 @@ def build_command:
...
+					error(\"expected only one manifests entry, not \" + (.manifests | length))
+				else . end
+
+				| .manifests[0] |= (
+					# https://github.com/docker-library/bashbrew/blob/4e0ea8d8aba49d54daf22bd8415fabba65dc83ee/cmd/bashbrew/oci-builder.go#L135-L144
+					if .mediaType != \"application/vnd.oci.image.manifest.v1+json\" then
+						error(\"unsupported descriptor mediaType: \" + .mediaType)
+					else . end
+					# TODO validate .digest somehow (`crane validate`? see below) - would also be good to validate all descriptors recursively (not sure if `crane push` does that)
+					| if .size < 0 then
```
whalelines, line 321: Is a size of zero valid?

whalelines (approved):
LGTM, a couple questions for my own edification

.test/library/ubuntu
```diff
@@ -0,0 +1,15 @@
+# https://github.com/docker-library/official-images/blob/fe9c059402181390eac083cbdd7229b5d123236e/library/ubuntu but intentionally slimmed down (just "latest" on one architecture, no email addresses)
+
+Maintainers: Tomáš Virtus (@woky), Cristóvão Cordeiro (@cjdcordeiro)
+GitRepo: https://git.launchpad.net/cloud-images/+oci/ubuntu-base
+GitCommit: fa42be9027eccb928a1f0f43d95ffd9a45d36737
+Builder: oci-import
+File: index.json
```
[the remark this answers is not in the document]
tianon: The `oci-import` builder is a bashbrew construction that uses OCI layouts, but does not mandate them explicitly -- see https://github.com/docker-library/bashbrew/blob/5152c0df682515cbe7ac62b68bcea4278856429f/cmd/bashbrew/oci-builder.go#L117-L134

If `File:` is not set to the literal string `index.json`, then it is assumed to point to a JSON file that contains a single OCI content descriptor for the image manifest (which will then be looked up under `blobs/sha256/xxxx`).

See also https://github.com/docker-library/meta-scripts/pull/20/files#diff-a64244565ecee6714e591bafe1b0771f08c5217287fb88a6ff0e9f1457eb8ac3R296-R299 for where in this PR those get "upgraded" into a correct `index.json` so that we don't have to handle both cases exceptionally here.

(_Technically_, the code I've written there to "upgrade" that case to `index.json` would break if the object under `blobs/sha256/xxx` is a symlink to `index.json` at the root of the OCI layout, but given we've established that the format is "OCI layout", it's fair for us to assume that `index.json` at the root has special meaning and whatever is happening in that theoretical case is wrong.)

meta.jq
```diff
@@ -265,10 +265,76 @@ def build_command:
...
+			"git init --bare \"$gitCache\"",
+			"_git() { git -C \"$gitCache\" \"$@\"; }",
+			"_git config gc.auto 0",
+			# "bashbrew fetch" but in Bash (because we have bashbrew, but not the library file -- we could synthesize a library file instead, but six of one half a dozen of another)
+			@sh "_commit() { _git rev-parse \(.source.entry.GitCommit + "^{commit}"); }",
+			@sh "if ! _commit &> /dev/null; then _git fetch \(.source.entry.GitRepo) \(.source.entry.GitCommit + ":") || _git fetch \(.source.entry.GitFetch + ":"); fi",
+			"_commit",
+
+			# TODO figure out a good, safe place to store our temporary build/push directory (maybe this is fine? we do it for buildx build too)
+			"mkdir temp",
```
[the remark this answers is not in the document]
tianon: It is, but `build_command` and `push_command` both need access to it, so it needs to be stored in a common location.  I think this really is fine as-is -- we've so far used the model that PWD is fair game for these commands to all store/transmit state, and it works reasonably well.

meta.jq
tianon: It's valid for the descriptor format, but in this case it would mean the image manifest itself is zero bytes, which _isn't_ valid.

.test/builds.json
```diff
@@ -2716,5 +2716,61 @@
         }
       }
     }
+  },
+  "93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf": {
+    "buildId": "93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf",
+    "build": {
+      "img": "oisupport/staging-amd64:93476ae64659d71f4ee7fac781d6d1890df8926682e2fa6bd647a246b33ad9bf",
+      "resolved": null,
```
tianon, line 2724: With https://github.com/docker-library/official-images/commit/795e04932a2ed061d24262c83560ab5a4e08d2ef, this is now even safer as part of our tests because this exact `buildId` will never be officially built. :smile: :innocent:

yosifkit (approved):

PR #23, opened 2024-02-16
Sort trigger-arch build queue to push previously failed jobs to the end

yosifkit:
This will give unbuilt items a chance to build before others get a retry.

(did a few successful run tests via Jenkins `trigger-arm32v6` job)

### DIFF
--- a/Jenkinsfile.trigger
+++ b/Jenkinsfile.trigger
@@ -13,6 +13,10 @@ env.BASHBREW_ARCH = env.JOB_NAME.split('/')[-1].minus('trigger-') // "windows-am
 def queue = []
 def breakEarly = false // thanks Jenkins...
 
+// this includes the number of attempts per failing buildId
+// { buildId: { "count": 1, ... }, ... }
+def pastFailedJobsJson = '{}'
+
 node {
 	stage('Checkout') {
 		checkout(scmGit(
@@ -32,33 +36,49 @@ node {
 				[$class: 'RelativeTargetDirectory', relativeTargetDir: 'meta'],
 			],
 		))
+		pastFailedJobsJson = sh(returnStdout: true, script: '''#!/usr/bin/env bash
+			set -Eeuo pipefail -x
+
+			if ! json="$(wget -qO- "$JOB_URL/lastSuccessfulBuild/artifact/pastFailedJobs.json")"; then
+				echo >&2 'failed to get pastFailedJobs.json'
+				json='{}'
+			fi
+			jq <<<"$json" '.'
+		''').trim()
 	}
 
 	dir('meta') {
-		def queueJSON = ''
+		def queueJson = ''
 		stage('Queue') {
-			// TODO this job should export a JSON file that includes the number of attempts so far per failing buildId, and then this list should inject those values, initialize missing to 0, and sort by attempts so that failing builds always live at the bottom of the queue
-			queueJSON = sh(returnStdout: true, script: '''
-				jq -L.scripts '
-					include "meta";
-					[
-						.[]
-						| select(
-							needs_build
-							and (
-								.build.arch as $arch
-								| if env.BASHBREW_ARCH == "gha" then
-									[ "amd64", "i386", "windows-amd64" ]
-								else [ env.BASHBREW_ARCH ] end
-								| index($arch)
+			withEnv([
+				'pastFailedJobsJson=' + pastFailedJobsJson,
+			]) {
+				// using pastFailedJobsJson, sort the needs_build queue so that failing builds always live at the bottom of the queue
+				queueJson = sh(returnStdout: true, script: '''
+					jq -L.scripts '
+						include "meta";
+						(env.pastFailedJobsJson | fromjson) as $pastFailedJobs
+						| [
+							.[]
+							| select(
+								needs_build
+								and (
+									.build.arch as $arch
+									| if env.BASHBREW_ARCH == "gha" then
+										[ "amd64", "i386", "windows-amd64" ]
+									else [ env.BASHBREW_ARCH ] end
+									| index($arch)
+								)
 							)
-						)
-					]
-				' builds.json
-			''').trim()
+						]
+						# this Jenkins job exports a JSON file that includes the number of attempts so far per failing buildId so that this can sort by attempts which means failing builds always live at the bottom of the queue (sorted by the number of times they have failed, so the most failing is always last)
+						| sort_by($pastFailedJobs[.buildId].count // 0)
+					' builds.json
+				''').trim()
+			}
 		}
-		if (queueJSON && queueJSON != '[]') {
-			queue = readJSON(text: queueJSON)
+		if (queueJson && queueJson != '[]') {
+			queue = readJSON(text: queueJson)
 			currentBuild.displayName = 'queue size: ' + queue.size() + ' (#' + currentBuild.number + ')'
 		} else {
 			currentBuild.displayName = 'empty queue (#' + currentBuild.number + ')'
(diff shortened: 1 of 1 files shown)


Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
...
+	withEnv([
+		'newFailedJobsJson=' + newFailedJobsJson,
+	]) {
+		stage('Archive') {
+			dir('builds') {
+				deleteDir()
+				sh '''#!/usr/bin/env bash
+					set -Eeuo pipefail -x
+
+					jq '.' <<<"$newFailedJobsJson" | tee pastFailedJobs.json
```
whalelines, line 217: Should the string `pastFailedJobs.json` be abstracted into a string since it is used several places?

whalelines (commented):
Looks good, a few questions

Jenkinsfile.trigger
```diff
@@ -32,33 +36,47 @@ node {
...
-								| index($arch)
+			withEnv([
+				'pastFailedJobsJson=' + pastFailedJobsJson,
+			]) {
+				// using pastFailedJobsJson, sort the needs_build queue so that failing builds always live at the bottom of the queue
+				queueJson = sh(returnStdout: true, script: '''
+					jq -L.scripts '
+						include "meta";
+						(env.pastFailedJobsJson | fromjson) as $pastFailedJobs |
+						[
```
tianon, line 61:
```suggestion
						(env.pastFailedJobsJson | fromjson) as $pastFailedJobs
						| [
```

Jenkinsfile.trigger
```diff
@@ -32,33 +36,47 @@ node {
...
+										[ "amd64", "i386", "windows-amd64" ]
+									else [ env.BASHBREW_ARCH ] end
+									| index($arch)
+								)
 							)
-						)
-					]
-				' builds.json
-			''').trim()
+						] | sort_by($pastFailedJobs[.buildId].count // 0)
```
tianon, line 73: Just to make sure:
```console
$ jq -n '{} | .foo.bar // "baz"'
"baz"
```
:heart:

tianon: Oh, but maybe we should include some bit of my old `TODO` comment here, to remind us what this is doing / for?

```suggestion
						]
						# this Jenkins job exports a JSON file that includes the number of attempts so far per failing buildId so that this can sort by attempts which means failing builds always live at the bottom of the queue (sorted by the number of times they have failed, so the most failing is always last)
						| sort_by($pastFailedJobs[.buildId].count // 0)
```

Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
...
+				def c = 1
+				if (pastFailedJobs[buildObj.buildId]) {
+					c += pastFailedJobs[buildObj.buildId].count
+				}
+				// TODO maybe implement some amount of backoff? keep first url/endtime?
+				newFailedJobs[buildObj.buildId] = [
+					count: c,
+					identifier: identifier,
+					url: res.absoluteUrl,
+					endtime: (res.startTimeInMillis + res.duration) / 1000.0, // convert to seconds
```
tianon, line 195:
```suggestion
					endTime: (res.startTimeInMillis + res.duration) / 1000.0, // convert to seconds
```

Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
...
 			)
 			// TODO do something useful with "res.result" (especially "res.result != 'SUCCESS'")
-			// (maybe store "res.startTimeInMillis + res.duration" as endTime so we can implement some amount of backoff somehow?)
 			echo(res.result)
 			if (res.result != 'SUCCESS') {
+				def c = 1
+				if (pastFailedJobs[buildObj.buildId]) {
+					c += pastFailedJobs[buildObj.buildId].count
+				}
+				// TODO maybe implement some amount of backoff? keep first url/endtime?
```
tianon, line 190:
```suggestion
				// TODO maybe implement some amount of backoff? keep first url/endTime?
```

Jenkinsfile.trigger
tianon, line 217:
```suggestion
					jq <<<"$newFailedJobsJson" '.' | tee pastFailedJobs.json
```

Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
 				quietPeriod: 5, // seconds
 			)
 			// TODO do something useful with "res.result" (especially "res.result != 'SUCCESS'")
-			// (maybe store "res.startTimeInMillis + res.duration" as endTime so we can implement some amount of backoff somehow?)
 			echo(res.result)
 			if (res.result != 'SUCCESS') {
+				def c = 1
+				if (pastFailedJobs[buildObj.buildId]) {
+					c += pastFailedJobs[buildObj.buildId].count
```
tianon, line 188: Something to think about in the future / next round: if `count` isn't set for some reason (but the object exists), this will be `null` and we'll get an execption from `java.lang.Integer#plus` (_shouldn't_ matter for this because we're the only ones creating this file and it should _always_ be well-formed, but this will be a good reminder for if/when something goes wrong that this line might need to be more defensive against the inputs).

tianon (commented):
A few really minor nits (none of which actually change the functionality so shouldn't need rigorous re-testing), but overall really solid and seems like it'll work just fine :+1:

Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
...
+				deleteDir()
+				sh '''#!/usr/bin/env bash
+					set -Eeuo pipefail -x
+
+					jq '.' <<<"$newFailedJobsJson" | tee pastFailedJobs.json
+				'''
+				archiveArtifacts(
+					artifacts: '*.json',
+					fingerprint: true,
+					onlyIfSuccessful: true,
```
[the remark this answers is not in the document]
tianon: This confusingly refers to the success of _this_ job, not the success of all the jobs this job is triggering (nor all the stages of this job, which each have their own status).  This is a nice way to give us a safety against losing our data since we're doing a very poor-man's linked list here and one hiccup means we lose our intermediate data. :smile:

(Hence pulling the previous version of this file from the `lastSuccessful` instead of just the most recent job.)

Jenkinsfile.trigger
```diff
@@ -161,12 +181,47 @@ for (buildObj in queue) {
...
+		stage('Archive') {
+			dir('builds') {
+				deleteDir()
+				sh '''#!/usr/bin/env bash
+					set -Eeuo pipefail -x
+
+					jq '.' <<<"$newFailedJobsJson" | tee pastFailedJobs.json
+				'''
+				archiveArtifacts(
+					artifacts: '*.json',
```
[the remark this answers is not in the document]
tianon: :shrug: we know the directory is definitely 100% empty (`deleteDir()` above), so I'm not too worried about it either way

Jenkinsfile.trigger
tianon: Yeah, that could help cut down on typos, but it does get a little annoying to inject properly (we'd probably need to do something like `env.PAST_FAILED_JOBS_FILENAME = ...` or something so that it can be accessed from all the shell invocations without dealing with quoting hell).

tianon (approved):

tianon:
CI failure is #25, fixed in #26, and this file isn't tested by CI anyhow :+1:

Issue #25, opened 2024-02-16
Launchpad does not allow fetching arbitrary commits, which now breaks our tests here


tianon:
```
+ .test/test.sh
failed fetching git repo for "ubuntu" (tags "22.04, jammy-20240111, jammy, latest" on arch "amd64")
failed fetching "e6e3490ad3f[5](https://github.com/docker-library/meta-scripts/actions/runs/7936297386/job/21671258830?pr=24#step:5:6)24ccaa072edafe525f8ca8ac5490:refs/remotes/temp2901481880/temp": server does not support exact SHA1 refspec
failed fetching "refs/tags/dist-jammy-amd[6](https://github.com/docker-library/meta-scripts/actions/runs/7936297386/job/21671258830?pr=24#step:5:7)4-20240111-e6e3490a:refs/remotes/temp29014[8](https://github.com/docker-library/meta-scripts/actions/runs/7936297386/job/21671258830?pr=24#step:5:9)1880/temp": couldn't find remote ref "refs/tags/dist-jammy-amd64-20240111-e6e3490a"
```

We can fix this in a few different ways.  The easiest/quickest temporary fix would be to mirror those old commits to a GitHub repository somewhere and update our test library file to reference it from GitHub instead (I still have these commits locally, so that would be pretty easy -- choosing/naming a repo is much harder; we used to use https://github.com/docker-library/commit-warehouse for this, before GitHub added support for fetching arbitrary commits and before it grew too large to even use anymore :innocent:).

A better long-term fix would probably be to finally convert `busybox` over to use `oci-import` (now that we have #20) which would also allow us to _test_ #20 properly. :smile:

tianon:
Ok, I took both commits from https://github.com/docker-library/meta-scripts/blob/b446700b3ec0b10ef89204ab585ceb862f131da5/.test/library/ubuntu (`fa42be9027eccb928a1f0f43d95ffd9a45d36737` and `e6e3490ad3f524ccaa072edafe525f8ca8ac5490`) and pushed them both into https://github.com/tianon/temp:

- https://github.com/tianon/temp/commit/fa42be9027eccb928a1f0f43d95ffd9a45d36737
- https://github.com/tianon/temp/commit/e6e3490ad3f524ccaa072edafe525f8ca8ac5490

I'll make a PR fixing our test shortly.

whalelines:
Did the behavior of Launchpad change? Why is this test failing now but was not before?

tianon:
[...]

PR #26, opened 2024-02-16
Move Launchpad references to my temp GitHub for now

tianon:
In the future, we'll hopefully get `busybox` converted over to `oci-import` so that this will no longer be necessary and we can delete this whole file instead (in favor of a `busybox` file instead).

Fixes https://github.com/docker-library/meta-scripts/issues/25

### DIFF
--- a/.test/builds.json
+++ b/.test/builds.json
@@ -2737,7 +2737,7 @@
         "ubuntu:latest"
       ],
       "entry": {
-        "GitRepo": "https://git.launchpad.net/cloud-images/+oci/ubuntu-base",
+        "GitRepo": "https://github.com/tianon/temp.git",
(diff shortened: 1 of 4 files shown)


tianon:
Unfortunately, this _does_ mean that we're no longer verifying that our `oci-import` jobs do not accidentally hit our "missing `.git` suffix" error, but I don't think that tiny benefit is worth keeping this actively up-to-date with the latest Ubuntu releases.

https://github.com/docker-library/meta-scripts/blob/b446700b3ec0b10ef89204ab585ceb862f131da5/meta.jq#L75

whalelines (approved):
LGTM

yosifkit (approved):

PR #28, opened 2024-02-28
Add a 3 hour timeout to individual builds

tianon:
This was part of the old system, but we hadn't yet had the problems that require it resurface yet.  The limit here is kind of arbitrary, but the specific value isn't as important as having one -- we had a build hang for 6d (due to some tool that doesn't have a timeout of its own but arguably should), and this prevents that from stalling our whole process.

### DIFF
--- a/Jenkinsfile.build
+++ b/Jenkinsfile.build
@@ -84,65 +84,67 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			return
 		}
 
-		/*
-		// TODO this is currently already done on the worker machines themselves, which is a tradeoff
-		// make sure "docker login" is localized to this workspace
-		env.DOCKER_CONFIG = workspace + '/.docker'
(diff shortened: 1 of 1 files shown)


tianon:
(This is definitely one to review without whitespace: https://github.com/docker-library/meta-scripts/pull/28/files?w=1)

tianon:
Also relevant for our GHA builds: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

(They default to 360 minutes, which is double this limit we've been using in the past -- their hardware is also roughly half or even less as capable as ours, so it seems fair to leave that at the default for now unless/until we have a compelling reason to change it or even make it explicitly part of this repository so they're consistent. :shrug:)

whalelines (approved):
LGTM

LaurentGoderre (approved):

yosifkit (approved):

PR #29, opened 2024-03-06
Swap Ubuntu for BusyBox in our test suite

tianon:
This avoids the Launchpad issue (#25 / #26) and gives us something small enough that I don't feel bad about leaving in the non-amd64 architectures so we keep the original pristine upstream library file as-is.

(See https://github.com/docker-library/official-images/pull/16356 :rocket:)


yosifkit (approved):

yosifkit:
Just need to wait for the `busybox:latest` images to populate into `oisupport/staging-*` and then regenerate to fix CI

yosifkit (approved):

PR #32, opened 2024-03-21
Add a `Reference` wrapper type to `ociref.Reference` to handle our "Docker references" logic more cleanly

tianon:
This also allows us to add explicit JSON round-tripping which can handle normalizing/denormalizing for us, so our output strings no longer contain `docker.io/[library/]`. 🎉

### DIFF
--- a/.test/builds.json
+++ b/.test/builds.json
@@ -13,7 +13,7 @@
             "size": 2372,
             "annotations": {
               "com.docker.official-images.bashbrew.arch": "amd64",
-              "org.opencontainers.image.ref.name": "docker.io/oisupport/staging-amd64:4b199ac326c74b3058a147e14f553af9e8e1659abc29bd3e82c9c9807b66ee43@sha256:0432a4d379794811b4a2e01d0d3e67a9bcf95d6c2bf71545f03bce3f1d60f401",
+              "org.opencontainers.image.ref.name": "oisupport/staging-amd64:4b199ac326c74b3058a147e14f553af9e8e1659abc29bd3e82c9c9807b66ee43@sha256:0432a4d379794811b4a2e01d0d3e67a9bcf95d6c2bf71545f03bce3f1d60f401",
               "org.opencontainers.image.revision": "6d541d27b5dd12639e5a33a675ebca04d3837d74",
               "org.opencontainers.image.source": "https://github.com/docker-library/docker.git#6d541d27b5dd12639e5a33a675ebca04d3837d74:24/cli",
               "org.opencontainers.image.url": "https://hub.docker.com/_/docker",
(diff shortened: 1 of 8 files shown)


tianon:
Poor man's codecov.io:

```diff
-.../registry/ref.go:18:                ParseRefNormalized         87.5%
+.../registry/ref.go:16:                ParseRef                   83.3%
+.../registry/ref.go:32:                Normalize                  100.0%
+.../registry/ref.go:45:                String                     100.0%
+.../registry/ref.go:54:                MarshalText                0.0%
+.../registry/ref.go:59:                UnmarshalText              0.0%
-.../registry/synthesize-index.go:17:   SynthesizeIndex            74.1%
+.../registry/synthesize-index.go:16:   SynthesizeIndex            74.1%
-.../registry/synthesize-index.go:141:  setRefAnnotation           100.0%
+.../registry/synthesize-index.go:140:  setRefAnnotation           100.0%
-.../registry/synthesize-index.go:151:  normalizeManifestPlatform  84.6%
+.../registry/synthesize-index.go:150:  normalizeManifestPlatform  84.6%
-total:                                 (statements)               74.3%
+total:                                 (statements)               73.7%
```

(I _think_ the poor scores on `MarshalText` and `UnmarshalText` are due to quirks of how Go counts these things, because they're definitely tested :upside_down_face:)

tianon:
image

The ~5% drop in `ParseRefNormalized` vs `ParseRef` is because `Normalize`'s logic used to be in there, but now isn't, so the error handling case is now a larger percentage of the overall function body. :upside_down_face:

tianon:
> (I _think_ the poor scores on `MarshalText` and `UnmarshalText` are due to quirks of how Go counts these things, because they're definitely tested 🙃)

Ok, something is definitely wrong with these - they really _aren't_ being invoked, and I'm not actually sure why! :sob:

tianon:
Ok, managed to find both why it wasn't getting invoked when I thought it should be _and_ updated my test to test it explicitly. :metal:

tianon:
_Much_ better:

```diff
@@ -16,12 +16,16 @@
 .../registry/cache.go:92:              GetBlob                    100.0%
 .../registry/cache.go:96:              GetManifest                100.0%
 .../registry/cache.go:100:             GetTag                     69.6%
-.../registry/client.go:17:             Client                     53.1%
+.../registry/client.go:17:             Client                     55.1%
-.../registry/client.go:135:            EntryForRegistry           72.7%
+.../registry/client.go:135:            EntryForRegistry           81.8%
 .../registry/rate-limits.go:23:        Do                         50.0%
 .../registry/read-helpers.go:15:       readJSONHelper             60.9%
-.../registry/ref.go:18:                ParseRefNormalized         87.5%
+.../registry/ref.go:16:                ParseRef                   83.3%
+.../registry/ref.go:32:                Normalize                  100.0%
+.../registry/ref.go:45:                String                     100.0%
+.../registry/ref.go:54:                MarshalText                100.0%
+.../registry/ref.go:59:                UnmarshalText              100.0%
-.../registry/synthesize-index.go:17:   SynthesizeIndex            74.1%
+.../registry/synthesize-index.go:16:   SynthesizeIndex            77.6%
-.../registry/synthesize-index.go:141:  setRefAnnotation           100.0%
+.../registry/synthesize-index.go:140:  setRefAnnotation           100.0%
-.../registry/synthesize-index.go:151:  normalizeManifestPlatform  84.6%
+.../registry/synthesize-index.go:150:  normalizeManifestPlatform  84.6%
-total:                                 (statements)               74.3%
+total:                                 (statements)               75.6%
```

https://github.com/docker-library/meta-scripts/actions/runs/8382812520/job/22957329243?pr=32#step:5:426

(I don't know for sure why `EntryForRegistry` and `Client` have different coverage here, but I'm guessing we have some quirks on how GitHub is setting up our workers, especially WRT authentication :upside_down_face:)

yosifkit (approved):

tianon:
> (I don't know for sure why `EntryForRegistry` and `Client` have different coverage here, but I'm guessing we have some quirks on how GitHub is setting up our workers, especially WRT authentication 🙃)

I just figured out `EntryForRegistry`'s inconsistency, and I literally have a `TODO` comment in the code for it!! :facepalm:

https://github.com/docker-library/meta-scripts/blob/1456103cd51117eab5b4ced14aef1b12cd8d62f0/registry/client.go#L140-L144

Because our iteration order is random, sometimes we get coverage on this first `continue` and sometimes we do not! (and the function is small enough that this actually affects the overall percentage by like 9.1%!)

Thanks Gobama!!

PR #34, opened 2024-03-22
Update registry cache to (ab)use `Data` field of `Descriptor` objects

tianon:
This allows us to store full descriptors (necessary to implement `Resolve*`), but with a net decrease in the number of fields we have to juggle / keep in sync.

This does mean consumers need to be careful about how they use the `Descriptor` objects we return (esp. WRT `Data`), but it makes it easier for them to then have `Data` available if they want it (which is something I'd like to use in the future).  This is a net win anyhow because the upstream objects might've contained `Data` fields so this forces us to deal with them in a sane way we're comfortable with instead of potentially just including them verbatim unintentionally. 🚀

Also: Implement `Resolve{Manifest,Tag,Blob}` in our registry cache

Now that we have `Descriptor` objects (and a way to cache them), this implementation is trivial. 🎉

### DIFF
--- a/registry/cache.go
+++ b/registry/cache.go
@@ -134,4 +134,73 @@ func (rc *registryCache) GetTag(ctx context.Context, repo string, tag string) (o
 	return ocimem.NewBytesReader(desc.Data, desc), nil
 }
 
+func (rc *registryCache) resolveBlob(ctx context.Context, repo string, digest ociregistry.Digest, f func(ctx context.Context, repo string, digest ociregistry.Digest) (ociregistry.Descriptor, error)) (ociregistry.Descriptor, error) {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	if desc, ok := rc.data[digest]; ok && rc.has[cacheKeyDigest(repo, digest)] {
+		return desc, nil
+	}
+
+	desc, err := f(ctx, repo, digest)
+	if err != nil {
+		return desc, err
+	}
+
+	digest = desc.Digest // if this isn't a no-op, we've got a naughty registry
+
+	rc.has[cacheKeyDigest(repo, digest)] = true
+
+	// carefully copy only valid Resolve* fields such that any other existing fields are kept (this matters more if we ever make our mutexes better/less aggressive 👀)
+	if d, ok := rc.data[digest]; ok {
+		d.MediaType = desc.MediaType
+		d.Digest = desc.Digest
+		d.Size = desc.Size
+		desc = d
+	}
+	rc.data[digest] = desc
+
+	return desc, nil
+}
+
+func (rc *registryCache) ResolveManifest(ctx context.Context, repo string, digest ociregistry.Digest) (ociregistry.Descriptor, error) {
+	return rc.resolveBlob(ctx, repo, digest, rc.registry.ResolveManifest)
+}
+
+func (rc *registryCache) ResolveBlob(ctx context.Context, repo string, digest ociregistry.Digest) (ociregistry.Descriptor, error) {
+	return rc.resolveBlob(ctx, repo, digest, rc.registry.ResolveBlob)
+}
+
+func (rc *registryCache) ResolveTag(ctx context.Context, repo string, tag string) (ociregistry.Descriptor, error) {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	tagKey := cacheKeyTag(repo, tag)
+
+	if digest, ok := rc.tags[tagKey]; ok {
+		if desc, ok := rc.data[digest]; ok {
+			return desc, nil
+		}
+	}
+
+	desc, err := rc.registry.ResolveTag(ctx, repo, tag)
+	if err != nil {
+		return desc, err
+	}
+
+	rc.has[cacheKeyDigest(repo, desc.Digest)] = true
+	rc.tags[tagKey] = desc.Digest
+
+	// carefully copy only valid Resolve* fields such that any other existing fields are kept (this matters more if we ever make our mutexes better/less aggressive 👀)
+	if d, ok := rc.data[desc.Digest]; ok {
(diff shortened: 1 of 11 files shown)


tianon:
See https://github.com/cue-labs/oci/pull/29 for the upstream reference of my latest commit -- it fixes the ability to `HEAD` against Docker Hub blob digests (which don't return a header of the digest).

registry/cache.go
```diff
@@ -104,43 +99,108 @@ func (rc *registryCache) GetTag(ctx context.Context, repo string, tag string) (o
...
-		rc.data[desc.Digest] = b
-	} else {
-		delete(rc.data, desc.Digest)
+	rc.data[desc.Digest] = desc
+
+	return ocimem.NewBytesReader(desc.Data, desc), nil
+}
+
+func (rc *registryCache) resolveBlob(ctx context.Context, repo string, digest ociregistry.Digest, f func(ctx context.Context, repo string, digest ociregistry.Digest) (ociregistry.Descriptor, error)) (ociregistry.Descriptor, error) {
+	rc.mu.Lock()
```
whalelines, line 138: Why do we have to lock on reads?

whalelines (commented):
Is it possible to unit test the new functions?

registry/cache.go
```diff
@@ -52,41 +50,38 @@ func (rc *registryCache) getBlob(ctx context.Context, repo string, digest ocireg
...
 	desc := r.Descriptor()
+	digest = desc.Digest // if this isn't a no-op, we've got a naughty registry
 
-	rc.has[cacheKeyDigest(repo, desc.Digest)] = true
-	rc.types[desc.Digest] = desc.MediaType
+	rc.has[cacheKeyDigest(repo, digest)] = true
+
+	if desc.Size > manifestSizeLimit {
+		rc.data[digest] = desc
+		return r, nil
```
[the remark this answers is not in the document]
tianon: Correct -- that's in the previous implementation too, but handling it earlier turns out to be better because we can then return the original reader earlier instead of buffering it.

Also, "too big" here is currently 4MiB, so it's a pretty generous amount of things that will be cached in memory anyhow (especially since we're looking at mostly just indexes, image manifests, and config blobs).

registry/cache.go
tianon: Unfortunately it's unsafe for two goroutines to read a map otherwise (and because if this _does_ end up querying the remote registry, it needs to write also).  The "data is cached" implementation should be really fast, though, so there shouldn't be much lock contention here (the contention is going to come from actual "load-bearing" concurrent requests :disappointed:).

tianon:
> Is it possible to unit test the new functions?

It's not trivial to do so, but it is possible.  What I've been looking at instead is trying to adjust `cmd/lookup` so that we can invoke these other functions also and thus get more integration coverage (which isn't exactly the same, but still good and high quality coverage and much easier to write / write correctly).

tianon:
Dropping to draft while I wrangle the testing situation (might need to get #32 first and rebase so the conflicts aren't too awful :sweat_smile:).

tianon:
```diff
@@ -3,7 +3,7 @@
 .../cmd/builds/main.go            loadCacheFromFile          76.0%
 .../cmd/builds/main.go            saveCacheToFile            78.6%
 .../cmd/builds/main.go            main                       80.7%
-.../cmd/lookup/main.go            main                       76.9%
+.../cmd/lookup/main.go            main                       82.6%
 .../om/om.go                      Keys                       100.0%
 .../om/om.go                      Get                        100.0%
 .../om/om.go                      Set                        100.0%
@@ -12,20 +12,25 @@
 .../registry/cache.go             RegistryCache              100.0%
 .../registry/cache.go             cacheKeyDigest             100.0%
 .../registry/cache.go             cacheKeyTag                100.0%
-.../registry/cache.go             getBlob                    75.0%
+.../registry/cache.go             getBlob                    81.0%
 .../registry/cache.go             GetBlob                    100.0%
 .../registry/cache.go             GetManifest                100.0%
-.../registry/cache.go             GetTag                     69.6%
+.../registry/cache.go             GetTag                     82.6%
+.../registry/cache.go             resolveBlob                68.8%
+.../registry/cache.go             ResolveManifest            100.0%
+.../registry/cache.go             ResolveBlob                100.0%
+.../registry/cache.go             ResolveTag                 77.8%
-.../registry/client.go            Client                     55.1%
+.../registry/client.go            Client                     54.2%
-.../registry/client.go            EntryForRegistry           81.8%
+.../registry/client.go            EntryForRegistry           72.7%
+.../registry/lookup.go            Lookup                     87.1%
-.../registry/rate-limits.go       Do                         50.0%
+.../registry/rate-limits.go       RoundTrip                  42.9%
 .../registry/read-helpers.go      readJSONHelper             60.9%
 .../registry/ref.go               ParseRef                   83.3%
 .../registry/ref.go               Normalize                  100.0%
 .../registry/ref.go               String                     100.0%
 .../registry/ref.go               MarshalText                100.0%
 .../registry/ref.go               UnmarshalText              100.0%
-.../registry/synthesize-index.go  SynthesizeIndex            77.6%
+.../registry/synthesize-index.go  SynthesizeIndex            81.1%
 .../registry/synthesize-index.go  setRefAnnotation           100.0%
-.../registry/synthesize-index.go  normalizeManifestPlatform  84.6%
+.../registry/synthesize-index.go  normalizeManifestPlatform  83.9%
-total:                            (statements)               75.6%
+total:                            (statements)               76.9%
```

(https://github.com/docker-library/meta-scripts/actions/runs/8398001647/job/23002236946#step:5:426 vs https://github.com/docker-library/meta-scripts/actions/runs/8426913314/job/23076302317?pr=34#step:5:426)

tianon:
(re the `EntryForRegistry` variance, see #36)

yosifkit (approved):

PR #35, opened 2024-03-22
Enforce `gofmt` in GitHub Actions

tianon:


.github/workflows/ci.yml
```diff
@@ -36,5 +36,7 @@ jobs:
           name: coverage
           path: .test/coverage**
           if-no-files-found: error
+      - name: gofmt
+        run: find -name '*.go' -type f -exec ./.go-env.sh gofmt -l -s -w '{}' +
```
whalelines, line 40: Do you not need to supply a path to `find`?

whalelines (commented):
It would be nice if `gofmt` returned non-zero when using `-l` or `-d`.

.github/workflows/ci.yml
tianon: It defaults to the current working directory :+1:

tianon:
> It would be nice if `gofmt` returned non-zero when using `-l` or `-d`.

I think it _can_ be convinced to do so, but this way makes sure that our following `git diff` invocation catches any/all of them _and_ any other missed things from our "canonical JSON" (instead of fixing one error only to find another), so I think this might actually be slightly better for our case. :smile:

whalelines (approved):
LGTM

.github/workflows/ci.yml
whalelines: Must be a GNU find enhancement. BSD find errors.

tianon: Even BusyBox is happy with it! :sweat_smile: 

```console
$ docker run --rm busybox find -maxdepth 1 -name etc
./etc
```

whalelines: Should I read that as a plus or a minus?
/me ducks

yosifkit (approved):

PR #36, opened 2024-03-25
Fix `EntryForRegistry` code coverage randomization

tianon:
I hate this, but https://github.com/golang/go/issues/61900 might someday make it slightly cleaner.

See https://github.com/docker-library/meta-scripts/pull/32#issuecomment-2018950756:

> Because our iteration order is random, sometimes we get coverage on this first `continue` and sometimes we do not! (and the function is small enough that this actually affects the overall percentage by like 9.1%!)

### DIFF
--- a/registry/client.go
+++ b/registry/client.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"slices"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


tianon:
Nice, and because `""` sorts above everything else, we end up with the "good" case where the `continue` is included in our coverage. :+1:

tianon:
`github.com/docker-library/meta-scripts/registry/client.go:135:		EntryForRegistry	81.8%` (main)

vs

`github.com/docker-library/meta-scripts/registry/client.go:136:		EntryForRegistry	86.7%` (this PR)

tianon:
I _think_ it's only a ~5% increase (instead of ~9%) because of all the lines this adds lolsob

yosifkit (approved):

PR #38, opened 2024-04-02
Rewrite deploy to minimize API requests

### LINKED ISSUE #22: scaling issues in deploy

tianon:
This replaces our `crane cp`/`crane index append` based deploy workflow with more custom code that can generate the full index/manifest list for each tag in a fully deterministic and offline way (based on the data we've already got locally in `builds.json`), and code that can take that result and push it directly with minimal API requests, especially in the no-op case (where `crane index append` would make something like $2*(N+1)$ API requests to no-op, this new code should be only $1$ total per tag in the no-op case, _even_ if we later add a pre-flight `HEAD` request which this code currently skips in favor of directly doing `PUT` and handling errors as a miss).

This also includes new integration tests and even some new `jq` tests to cover the new code.  As noted in a comment in the code, in theory we can use `ocimem` in the future to add more unit tests to this code, but it's harder to successfully/correctly test the edge cases that way (since we'd then have to synthesize content in-process instead of relying on pre-synthesized content pre-existing), so I felt pretty good about leaving that as a `TODO` for now.

(I have been using this successfully for deploying my own personal images, FWIW. :+1:)

Closes #22


tianon:
This is real big.  For now, I've left it in "many WIP commits" mode, which might make it easier or more interesting to review, but I don't think we should merge all these commits (and I was planning to squash them).

tianon:
For purposes of comparison, the Go coverage percentage on `main` is currently 77.1% total, and this PR is at 77.9% total :smile:

Jenkinsfile.deploy
```diff
@@ -3,11 +3,14 @@ properties([
...
+		durationName: 'hour',
+		userBoost: true,
+	]),
 	pipelineTriggers([
-		// TODO https://github.com/docker-library/meta-scripts/issues/22
-		//upstream(threshold: 'UNSTABLE', upstreamProjects: 'meta'),
-		cron('H H/2 * * *'),
-		// (we've dropped to only running this periodically to avoid it clogging the whole queue for a no-op, which also gives build+meta more time to cycle and get deps so they have a higher chance to all go out at once -- see the above linked issue)
+		upstream('meta'),
+		cron('H H/6 * * *'), // run every few hours whether we "need" it or not
```
tianon, line 13: This effectively sets us to run `deploy` every time `meta` succeeds, but at most once per hour (unless we run manually).  I don't personally think this is _right_, but it's maybe close enough for first pass (triggers relatively frequently, but not so frequently that we cause problems for anyone else).

I'd also add the reminder that this job running is only a blocker for _users_ to see new builds, not for dependent/child builds, so it's pretty reasonable for them to have a 1-2 hour delay (and we can always run manually if we know the builds are done and need to get a deploy out faster than that).

cmd/deploy/input.go
```diff
@@ -0,0 +1,246 @@
...
+				return registry.CopyBlob(ctx, *normal.CopyFrom, dstRef)
+			}
+		}
+
+	default:
+		panic("unknown type: " + string(normal.Type))
+		// panic instead of error because this should've already been handled/normalized above (so this is a coding error, not a runtime error)
+	}
+
+	return normal, nil
```
whalelines, line 245: That is quite a function.

whalelines (approved):
Looks good, a couple questions

.test/oci-sort-platforms/test.jq
```diff
@@ -0,0 +1,72 @@
...
+	},
+
+	# buildkit attestations
+	# https://github.com/moby/buildkit/blob/5e0fe2793d529209ad52e811129f644d972ea094/docs/attestations/attestation-storage.md#attestation-manifest-descriptor
+	{
+		architecture: "unknown",
+		os: "unknown",
+	},
+
+	empty
```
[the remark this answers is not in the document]
tianon: Smaller diffs in the future -- it allows a "list" to end with a trailing comma (which is otherwise invalid syntax) so that we don't have to modify the previous item to add a new one :+1:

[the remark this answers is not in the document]
tianon: (sometimes I write this as `empty # trailing comma hack` but it's a little tedious to do so consistently :sweat_smile:)

cmd/deploy/input.go
tianon: Yeahhhhhhh -- I actually wrote this as part of `main()` itself, and when I stepped back from the implementation I almost died and thus forced myself to re-write it in such a way that I could load it up with unit tests so I could have higher confidence that all the edge cases in every branch of this behemoth are actually covered properly.

yosifkit (approved):

PR #39, opened 2024-04-19
breakup function NormalizeInput and add tests

LaurentGoderre:


cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+			"localhost:5000/foo:#: failed to parse ref: tag \"#\" does not start with word character",
+		},
+		{
+			"invalid reference url",
+			[]string{
+				"localhost:5000/foo?test",
+			},
+			typeManifest,
+			want{},
+			"localhost:5000/foo?test: failed to parse ref: tag \"5000/foo?test\" contains invalid invalid character '/'",
```
tianon, line 88: These are great examples of why testing the exact error message makes me nervous -- if Go decides (for whatever reason) to pull in a newer version of the `ociregistry` module, these error messages might all change because they ultimately come from that code, so we're effectively testing _their_ code instead of ours.  I don't know that there's a better answer here except maybe _only_ looking for `failed to parse ref` (our bit of the error) in the error message? (https://pkg.go.dev/strings#Contains)

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+			if len(lookups) != len(x.want.lookups) {
+				t.Errorf("Reference count mismatch. got: %d, expected: %d", len(lookups), len(x.want.lookups))
+			}
+
+			for digest, ref := range lookups {
+				if _, ok := x.want.lookups[digest]; !ok {
+					t.Errorf("digest missing from lookups: %s", digest)
+				} else if ref != x.want.lookups[digest] {
+					gotJson, _ := json.MarshalIndent(MarshallableReference(ref), "", "  ")
+					expectedJson, _ := json.MarshalIndent(MarshallableReference(x.want.lookups[digest]), "", "  ")
```
tianon, line 230: Even reading this, I'm still confused :sweat_smile: couldn't this just be `%+v` or `%#v` in the error message?

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+					t.Errorf("Reference doesn't match. got: %s, expected: %s", gotJson, expectedJson)
+				}
+			}
+			_ = lookups
```
tianon, line 234: :thinking: (probably just forgot to delete this?)

cmd/deploy/input.go
```diff
@@ -49,63 +49,46 @@ type inputNormalized struct {
...
-		normal.Refs[i] = ref
+		refs[i] = ref
 	}
-	debugId := normal.Refs[0]
 
-	normal.Lookup = make(map[ociregistry.Digest]registry.Reference, len(raw.Lookup))
+	return refsDigest, refs, nil
+}
+
+func getLookup(rawLookup map[string]string) (ociregistry.Digest, map[ociregistry.Digest]registry.Reference, error) {
```
tianon, line 79:
```suggestion
func normalizeInputLookup(rawLookup map[string]string) (map[ociregistry.Digest]registry.Reference, ociregistry.Digest, error) {
```

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+		t.Run(x.name, func(t *testing.T) {
+			refsDigest, refs, err := getRefs(x.rawDigests, x.deployType)
+
+			if x.wantErr != "" && (err == nil || err.Error() == "") {
+				t.Fatalf("Expected error not returned: %s", x.wantErr)
+			}
+
+			if x.wantErr != "" && err.Error() != x.wantErr {
+				t.Fatalf("Expected error doesn't match. got:\n'%s',\n\nexpected:\n'%s'", err.Error(), x.wantErr)
+			}
```
tianon, line 119: Checking whether we got an error object but that it returns the empty string for the error message is a bit strange -- I also think this code would be easier to follow if it had `if x.wantErr != ""` only once, and a sub-`if` for `if err == nil { fatal(unexpected errorless return) }` (which should maybe also include the other things we _did_ get in the error so it's easier to narrow down where it's failing) + something like `if errStr := err.Error(); !strings.Contains(errStr, x.wantErr) { fatal(unexpected err (looking for %q): errStr, x.wantErr) }`

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+
+			if refsDigest != x.want.digest {
+				t.Errorf("got:\n%s\n\nexpected:\n%s\n", string(refsDigest), x.want.digest)
+			}
+
+			if len(refs) != len(x.want.refs) {
+				t.Errorf("Reference count mismatch. got: %d, expected: %d", len(refs), len(x.want.refs))
+			}
```
tianon, line 127: I think this is probably overkill -- we should just let the slice comparison handle this instead :+1:

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+
+			if !reflect.DeepEqual(refs, x.want.refs) {
+				gotJson, _ := json.MarshalIndent(refs, "", "  ")
+				expectedJson, _ := json.MarshalIndent(x.want.refs, "", "  ")
+				t.Errorf("Reference doesn't match. got: %s, expected: %s", gotJson, expectedJson)
```
tianon, line 132: Ah, we don't need JSON here -- probably just `%+v` or `%#v`

tianon (commented):
Just to record here what we talked about offline, I wasn't originally sold on the idea of splitting this up without more consumers of the split functions or without the flow being obviously cleaner, but it does make testing the error cases more straightforward, so I'm convinced. :+1:

I've left a few notes (many of which imply more changes than just the lines they're on :sweat_smile:). :+1:

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+	}
+}
+
+func TestGetLookup(t *testing.T) {
+	type MarshallableReference struct {
+		Host       string             `json:"Host"`
+		Repository string             `json:"Repository"`
+		Tag        string             `json:"Tag"`
+		Digest     ociregistry.Digest `json:"Digest"`
+	}
```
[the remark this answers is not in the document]
LaurentGoderre: Only needed if we do JSON compare. That's the only reason I added this

cmd/deploy/input.go
```diff
@@ -115,13 +98,50 @@ func NormalizeInput(raw inputRaw) (inputNormalized, error) {
...
+		return normal, fmt.Errorf("unknown type: %s", raw.Type)
+	}
+
+	if raw.Refs == nil {
+		return normal, fmt.Errorf("missing refs entirely (JSON input glitch?)")
+	}
+	if len(raw.Refs) == 0 {
+		return normal, fmt.Errorf("zero refs specified for pushing (need at least one)")
+	}
+	refsDigest, refs, err := getRefs(raw.Refs, normal.Type)
```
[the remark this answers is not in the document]
LaurentGoderre: Huh!! I didn't know you could do that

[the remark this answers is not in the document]
LaurentGoderre: It wont compile: `non-name normal.Refs on left side of :=`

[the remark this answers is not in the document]
whalelines: You could declare `refsDigest` in a prior statement and then use a simple `=`.

cmd/deploy/input_test.go
LaurentGoderre: Nerat, though the pretty print feature of JSON helped readability

cmd/deploy/input.go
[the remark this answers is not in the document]
tianon: Doh, forgot Go is particular about `:=` like that -- yeah, adding a `var` is probably the better answer then:

```suggestion
	var refsDigest ociregistry.Digest
	normal.Refs, refsDigest, err = normalizeInputRefs(raw.Refs, normal.Type)
```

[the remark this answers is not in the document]
LaurentGoderre: I think the issue is this: https://github.com/golang/go/issues/30318

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,240 @@ package main
...
+		refs   []registry.Reference
+	}
+	type fixture struct {
+		name       string
+		rawDigests []string
+		deployType deployType
+		want       want
+		wantErr    string
+	}
+	for _, x := range []fixture{
```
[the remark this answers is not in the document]
LaurentGoderre: If I don't specify the `want` type explicitly I get `missing type in composite literal`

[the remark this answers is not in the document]
LaurentGoderre: There are a few different routes and I found this one to be the most compact and readable.

1) I could repeat the struct everywhere
2) I can flatten the structure but that gets tricky for testing error when you are not expecting output
3) the solution I chosen is the explicit type which allows to use `want want` in the fixture type and use `want{}` when testing for the error.

If there is another way in Go that I don't know, id be happy to learn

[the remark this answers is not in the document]
tianon: The solution is naming the fields :smile:

<details>
<summary>Patch:</summary>

```patch
diff --git a/cmd/deploy/input_test.go b/cmd/deploy/input_test.go
index e196ef5..2758d46 100644
--- a/cmd/deploy/input_test.go
+++ b/cmd/deploy/input_test.go
@@ -11,105 +11,92 @@ import (
 )
 
 func TestNormalizeInputRefs(t *testing.T) {
-	type want struct {
-		refs   []registry.Reference
-		digest ociregistry.Digest
-	}
-	type fixture struct {
+	for _, x := range []struct {
 		name       string
 		deployType deployType
-		rawDigests []string
-		want       want
+		rawRefs    []string
+		wantRefs   []registry.Reference
+		wantDigest ociregistry.Digest
 		wantErr    string
-	}
-	for _, x := range []fixture{
+	}{
 		{
-			"happy path",
-			typeManifest,
-			[]string{
+			name:       "happy path",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"localhost:5000/foo@sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
 				"localhost:5000/bar:some-tag",
 				"localhost:5000/baz",
 			},
-			want{
-				[]registry.Reference{
-					{
-						Host:       "localhost:5000",
-						Repository: "foo",
-						Digest:     "sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
-					},
-					{
-						Host:       "localhost:5000",
-						Repository: "bar",
-						Tag:        "some-tag",
-					},
-					{
-						Host:       "localhost:5000",
-						Repository: "baz",
-					},
+			wantRefs: []registry.Reference{
+				{
+					Host:       "localhost:5000",
+					Repository: "foo",
+					Digest:     "sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
+				},
+				{
+					Host:       "localhost:5000",
+					Repository: "bar",
+					Tag:        "some-tag",
+				},
+				{
+					Host:       "localhost:5000",
+					Repository: "baz",
 				},
-				"sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
 			},
-			"",
+			wantDigest: "sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
 		},
 		{
-			"invalid digest",
-			typeManifest,
-			[]string{
+			name:       "invalid digest",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"localhost:5000/foo@bad_digest",
 			},
-			want{},
-			"failed to parse ref:",
+			wantErr: "failed to parse ref:",
 		},
 		{
-			"invalid host",
-			typeManifest,
-			[]string{
+			name:       "invalid host",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"...",
 			},
-			want{},
-			"failed to parse ref:",
+			wantErr: "failed to parse ref:",
 		},
 		{
-			"invalid tag",
-			typeManifest,
-			[]string{
+			name:       "invalid tag",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"localhost:5000/foo:#",
 			},
-			want{},
-			"failed to parse ref:",
+			wantErr: "failed to parse ref:",
 		},
 		{
-			"invalid reference url",
-			typeManifest,
-			[]string{
+			name:       "invalid reference url",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"localhost:5000/foo?test",
 			},
-			want{},
-			"failed to parse ref:",
+			wantErr: "failed to parse ref:",
 		},
 		{
-			"mismatch digest",
-			typeManifest,
-			[]string{
+			name:       "mismatch digest",
+			deployType: typeManifest,
+			rawRefs: []string{
 				"localhost:5000/foo@sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
 				"localhost:5000/foo@sha256:00042f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
 			},
-			want{},
-			"refs digest mismatch in ",
+			wantErr: "refs digest mismatch in ",
 		},
 		{
-			"can't push to blob",
-			typeBlob,
-			[]string{
+			name:       "can't push to blob",
+			deployType: typeBlob,
+			rawRefs: []string{
 				"localhost:5000/foo:tag",
 			},
-			want{},
-			"cannot push blobs to a tag:",
+			wantErr: "cannot push blobs to a tag:",
 		},
 	} {
 		t.Run(x.name, func(t *testing.T) {
-			refs, refsDigest, err := normalizeInputRefs(x.deployType, x.rawDigests)
+			refs, refsDigest, err := normalizeInputRefs(x.deployType, x.rawRefs)
 
 			if x.wantErr != "" {
 				if err == nil {
@@ -117,16 +104,16 @@ func TestNormalizeInputRefs(t *testing.T) {
 				}
 
 				if !strings.Contains(err.Error(), x.wantErr) {
-					t.Fatalf("Expected error doesn't match. got:\n'%s',\n\nexpected to contain:\n'%s'", err.Error(), x.wantErr)
+					t.Fatalf("Expected error doesn't match. got:\n%v,\n\nexpected to contain:\n%q", err, x.wantErr)
 				}
 			}
 
-			if refsDigest != x.want.digest {
-				t.Errorf("got:\n%s\n\nexpected:\n%s\n", string(refsDigest), x.want.digest)
+			if refsDigest != x.wantDigest {
+				t.Errorf("got:\n%s\n\nexpected:\n%s\n", string(refsDigest), x.wantDigest)
 			}
 
-			if !slices.Equal(refs, x.want.refs) {
-				t.Errorf("Reference doesn't match. got: \n%#v,\n\nexpected:\n%#v", refs, x.want.refs)
+			if !slices.Equal(refs, x.wantRefs) {
+				t.Errorf("Reference doesn't match. got: \n%#v,\n\nexpected:\n%#v", refs, x.wantRefs)
 			}
 		})
 	}
```

</details>

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,234 @@ package main
...
+						Digest:     "sha256:0cb474919526d040392883b84e5babb65a149cc605b89b117781ab94e88a5e86",
+					},
+					"sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e": {
+						Host:       "localhost:5000",
+						Repository: "example",
+						Tag:        "bar",
+						Digest:     "sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e",
+					},
+				},
+				"sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e",
```
tianon, line 176: I think this might actually be a bug in the implementation -- maps are not ordered (in fact, Go actively randomizes them, which is why this test fails randomly), but in this case the function cannot have determined which value is the "target" or "common" digest, so it should be returning empty :thinking:

(catching this is a nice win from splitting, I think! :grinning:)

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,234 @@ package main
...
+			},
+			"",
+		},
+		{
+			"bad digest",
+			map[string]string{
+				"invalid digest": "example:foo",
+			},
+			want{},
+			"lookup key \"invalid digest\" invalid:",
```
tianon, line 186: If you swap this to use backticks, you don't have to escape the `"`:

```suggestion
			`lookup key "invalid digest" invalid:`,
```

cmd/deploy/input_test.go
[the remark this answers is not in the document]
tianon: To be clear: I'm happy to commit/push that patch somewhere if you'd like to take it as-is, or you can apply it locally yourself :+1:

cmd/deploy/input_test.go
tianon:
```go
	var commonDigest ociregistry.Digest // if we store this out here, we can abuse it later to get the "last" lookup digest (for getting the single key in the case of len(lookup) == 1 without a new loop)
```

aha -- so we should update the function to explicitly clear this value before returning it unless the length of our final return map is exactly 1

then we can _also_ update the code calling this function to use whether or not this value is empty instead of `len(lookup) == 1` to determine whether or not to use it :rocket:

tianon: Oh wait, empty might mean fallback reference, so more of that later conditional needs to move too :thinking:

tianon: This actually complicates it quite a bit more.  This function signature might need more thought. :thinking:

(Or we adjust the test, but it feels wrong to have the function return data that's not valid to use _and_ it feels wrong to copy so much conditional logic into the test.)

tianon: Ok, maybe something like this: (the new test needs updates for both this and the other thread, but at least this fixes the implementation)

<details>
<summary>Patch:</summary>

```patch
diff --git a/cmd/deploy/input.go b/cmd/deploy/input.go
index 2a2a736..1477e77 100644
--- a/cmd/deploy/input.go
+++ b/cmd/deploy/input.go
@@ -76,43 +76,48 @@ func normalizeInputRefs(deployType deployType, rawRefs []string) ([]registry.Ref
 	return refs, commonDigest, nil
 }
 
-func normalizeInputLookup(rawLookup map[string]string) (map[ociregistry.Digest]registry.Reference, ociregistry.Digest, error) {
-	lookups := make(map[ociregistry.Digest]registry.Reference, len(rawLookup))
-	var commonDigest ociregistry.Digest // if we store this out here, we can abuse it later to get the "last" lookup digest (for getting the single key in the case of len(lookup) == 1 without a new loop)
+func normalizeInputLookup(rawLookup map[string]string) (map[ociregistry.Digest]registry.Reference, *ociregistry.Digest, error) {
+	lookup := make(map[ociregistry.Digest]registry.Reference, len(rawLookup))
+	var digest ociregistry.Digest // if we store this out here, we can abuse it later to get the "last" lookup digest (for getting the single key in the case of len(lookup) == 1 without a new loop)
 	for d, refString := range rawLookup {
-		commonDigest = ociregistry.Digest(d)
-		if commonDigest != "" {
+		digest = ociregistry.Digest(d)
+		if digest != "" {
 			// normal.Lookup[""] is a special case for fallback (where to look for any child object that isn't explicitly referenced)
-			if err := commonDigest.Validate(); err != nil {
-				return nil, "", fmt.Errorf("lookup key %q invalid: %w", commonDigest, err)
+			if err := digest.Validate(); err != nil {
+				return nil, nil, fmt.Errorf("lookup key %q invalid: %w", digest, err)
 			}
 		}
 		if ref, err := registry.ParseRef(refString); err != nil {
-			return nil, "", fmt.Errorf("failed to parse lookup ref %q: %v", refString, err)
+			return nil, nil, fmt.Errorf("failed to parse lookup ref %q: %v", refString, err)
 		} else {
-			if ref.Tag != "" && commonDigest != "" {
+			if ref.Tag != "" && digest != "" {
 				//return normal, fmt.Errorf("%s: tag on by-digest lookup ref makes no sense: %s (%s)", debugId, ref, d)
 			}
 
-			if ref.Digest == "" && commonDigest != "" {
-				ref.Digest = commonDigest
+			if ref.Digest == "" && digest != "" {
+				ref.Digest = digest
 			}
-			if commonDigest != "" && ref.Digest != commonDigest {
-				return nil, "", fmt.Errorf("digest on lookup ref should either be omitted or match key: %s vs %s", ref, d)
+			if digest != "" && ref.Digest != digest {
+				return nil, nil, fmt.Errorf("digest on lookup ref should either be omitted or match key: %s vs %s", ref, d)
 			}
 
-			lookups[commonDigest] = ref
+			lookup[digest] = ref
 		}
 	}
 
-	return lookups, commonDigest, nil
+	// see notes on "digest" definition
+	if len(lookup) != 1 {
+		return lookup, nil, nil
+	}
+	if digest == "" && (lookup[""].Digest == "" && lookup[""].Tag == "") {
+		// if it was a fallback, it needs at least Tag or Digest (or our refs need Digest, so we can infer)
+		return lookup, nil, nil
+	}
+	return lookup, &digest, nil
 }
 
 func NormalizeInput(raw inputRaw) (inputNormalized, error) {
 	var normal inputNormalized
-	var lookupDigest ociregistry.Digest
-	var refsDigest ociregistry.Digest
-	var err error
 
 	switch raw.Type {
 	case "":
@@ -132,12 +137,18 @@ func NormalizeInput(raw inputRaw) (inputNormalized, error) {
 	if len(raw.Refs) == 0 {
 		return normal, fmt.Errorf("zero refs specified for pushing (need at least one)")
 	}
+	var (
+		refsDigest ociregistry.Digest
+		err error
+	)
 	normal.Refs, refsDigest, err = normalizeInputRefs(normal.Type, raw.Refs)
 	if err != nil {
 		return normal, err
 	}
 
-	debugId := normal.Refs[0]
+	debugId := normal.Refs[0] // used for annotating errors from here out
+
+	var lookupDigest *ociregistry.Digest
 	normal.Lookup, lookupDigest, err = normalizeInputLookup(raw.Lookup)
 	if err != nil {
 		return normal, fmt.Errorf("%s: %w", debugId, err)
@@ -147,29 +158,24 @@ func NormalizeInput(raw inputRaw) (inputNormalized, error) {
 		// if we have no Data, let's see if we have enough information to infer an object to copy
 		if lookupRef, ok := normal.Lookup[refsDigest]; refsDigest != "" && ok {
 			// if any of our Refs had a digest, *and* we have a way to Lookup that digest, that's the one
-			lookupDigest = refsDigest
+			lookupDigest = &refsDigest
 			normal.CopyFrom = &lookupRef
-		} else if lookupRef, ok := normal.Lookup[lookupDigest]; len(normal.Lookup) == 1 && ok {
+		} else if lookupDigest != nil {
 			// if we only had one Lookup entry, that's the one
-			if lookupDigest == "" {
-				// if it was a fallback, it needs at least Tag or Digest (or our refs need Digest, so we can infer)
-				if lookupRef.Digest == "" && lookupRef.Tag == "" {
-					if refsDigest != "" {
-						lookupRef.Digest = refsDigest
-					} else {
-						return normal, fmt.Errorf("%s: (single) fallback needs digest or tag: %s", debugId, lookupRef)
-					}
-				}
-			}
+			lookupRef := normal.Lookup[*lookupDigest]
+			normal.CopyFrom = &lookupRef
+		} else if lookupRef, ok := normal.Lookup[""]; refsDigest != "" && ok {
+			lookupDigest = &refsDigest
+			lookupRef.Digest = refsDigest
 			normal.CopyFrom = &lookupRef
 		} else {
 			// if Lookup has only a single entry, that's the one (but that's our last chance for inferring intent)
-			return normal, fmt.Errorf("%s: missing data (and lookup is not a single item)", debugId)
+			return normal, fmt.Errorf("%s: missing data (and lookup is not a single item or fallback with digest or tag)", debugId)
 			// TODO *technically* it would be fair to have lookup have two items if one of them is the fallback reference, but it doesn't really make much sense to copy an object from one namespace, but to get all its children from somewhere else
 		}
 
-		if lookupDigest == "" && normal.CopyFrom.Digest != "" {
-			lookupDigest = normal.CopyFrom.Digest
+		if *lookupDigest == "" && normal.CopyFrom.Digest != "" {
+			lookupDigest = &normal.CopyFrom.Digest
 		}
 
 		if _, ok := normal.Lookup[""]; !ok {
@@ -178,8 +184,8 @@ func NormalizeInput(raw inputRaw) (inputNormalized, error) {
 		}
 
 		if refsDigest == "" {
-			refsDigest = lookupDigest
-		} else if lookupDigest != "" && refsDigest != lookupDigest {
+			refsDigest = *lookupDigest
+		} else if *lookupDigest != "" && refsDigest != *lookupDigest {
 			return normal, fmt.Errorf("%s: copy-by-digest mismatch: %s vs %s", debugId, refsDigest, normal.CopyFrom)
 		}
 	} else {
```

</details>

cmd/deploy/input_test.go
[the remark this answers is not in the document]
LaurentGoderre: Nah, good for my learning to apply it myself

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,249 @@ package main
...
+				}
+
+				if !strings.Contains(err.Error(), x.wantErr) {
+					t.Fatalf("Expected error doesn't match. got:\n'%s',\n\nexpected to contain:\n'%s'", err.Error(), x.wantErr)
+				}
+			}
+
+			if x.wantDigest != "" {
+				if lookupDigest == nil {
+					t.Errorf("Expected last digest. got:\n%s\n\nexpected:\n%s\n", "<nil>", x.wantDigest)
```
tianon, line 231: I think it might be worth adding a special case like `wantDigest: "empty"` or something so we can test the edge cases around that (non-nil-but-empty-as-expected) -- single `lookup` entry with an empty key and value with a tag, value with a digest, value with both tag and digest? :eyes:

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,249 @@ package main
...
+		t.Run(x.name, func(t *testing.T) {
+			lookups, lookupDigest, err := normalizeInputLookup(x.rawLookups)
+
+			if x.wantErr != "" {
+				if err == nil {
+					t.Fatalf("Expected error not returned: %s", x.wantErr)
+				}
+
+				if !strings.Contains(err.Error(), x.wantErr) {
+					t.Fatalf("Expected error doesn't match. got:\n'%s',\n\nexpected to contain:\n'%s'", err.Error(), x.wantErr)
```
tianon, line 225: Same notes as similar line in previous function:

```suggestion
					t.Fatalf("Expected error doesn't match. got:\n%v\n\nexpected to contain: %q", err, x.wantErr)
```

tianon (commented):
Nice!  A few more comments, but really only minor nits in the tests and a few more edges I think we should cover. :+1:

(I think there might be a few cases of inconsistent "trailing newline" in `t.Errorf` calls, but I didn't go annotate every one of them because they don't bother me as much -- if you feel so inclined, I'd make those all consistently no-trailing-newline :eyes:)

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,249 @@ package main
...
+				}
+			}
+
+			for digest, ref := range lookups {
+				if _, ok := x.wantLookups[digest]; !ok {
+					t.Errorf("digest missing from lookups: %s", digest)
+				} else if ref != x.wantLookups[digest] {
+					t.Errorf("Reference doesn't match. got:\n%#v,\n\nexpected:\n%#v", ref, x.wantLookups[digest])
+				}
+			}
```
tianon, line 243: Oh, this might also miss things (like returning too much, for example) -- I think you probably want to use https://pkg.go.dev/maps#Equal here :eyes:

cmd/deploy/input_test.go
LaurentGoderre: Like this but more cases? https://github.com/docker-library/meta-scripts/pull/39/files#diff-51189177e2def5d122681ee1eecba1fec82822afe73d8fd24e6a861ca238773cR179-R190

tianon: yeah, specifically something like `{"": "example@sha256:xxx"}` (which is then an expected digest of `sha256:xxx`) and `{"": "example:tag"}` (which is an expected digest of explicitly `""` vs `nil` to signal that the fallback has a tag we should use)

LaurentGoderre: I added tests to match the current behavior but I'm not sure it matches what you said the intended behavior is

LaurentGoderre: I got it working. It was the test that was wrong.

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+		t.Run(x.name, func(t *testing.T) {
+			refs, refsDigest, err := normalizeInputRefs(x.deployType, x.rawRefs)
+
+			if x.wantErr != "" {
+				if err == nil {
+					t.Fatalf("Expected error not returned: %s", x.wantErr)
+				}
+
+				if !strings.Contains(err.Error(), x.wantErr) {
+					t.Fatalf("Expected error doesn't match.\ngot:\n%q,\n\nexpected to contain:\n%q", err, x.wantErr)
```
whalelines, line 109: Perhaps https://github.com/stretchr/testify would make all these error reports simpler.

whalelines (commented):
Thanks for the tests, one suggestion

cmd/deploy/input_test.go
LaurentGoderre: @whalelines I kind of agree but I started a thread on testing frameworks in one of our channel and it seems go developers tend to not want to use such framework. I will let @tianon decide

whalelines: The signing team is using testify to good effect.

whalelines: In most cases, less is code is better. The less code, the less to maintain. The best refactor is delete.

LaurentGoderre: im all for it if @tianon and @yosifkit are good with it. Although, could we perhaps merge first and refactor after? Lots of activity in this PR already

tianon: "A little copying is better than a little dependency." (https://go-proverbs.github.io/)

In this case, I don't think testify counts as a "little" dependency, but that's also often a good reason to avoid it -- generally in Go, dependencies are heavy, and create extra load on consumers of your code too.

I'm not opposed to `testify`, but the amount it makes our tests nicer needs to be greater than the amount of burden it (and its dependencies) place on us.  It looks like the dep tree for testify itself is pretty small, so it's probably reasonable, but I agree it's something to consider as a separate PR.

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+			name:       "mismatch digest",
+			deployType: typeManifest,
+			rawRefs: []string{
+				"localhost:5000/foo@sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
+				"localhost:5000/foo@sha256:00042f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
+			},
+			wantErr: "refs digest mismatch in ",
+		},
+		{
+			name:       "can't push to blob",
```
tianon, line 92:
```suggestion
			name:       "can't push blob to tag",
```

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+				"": {
+					Host:       "docker.io",
+					Repository: "library/example",
+				},
+			},
+			wantDigest: nil,
+			wantErr:    "",
+		},
+		{
+			name: "Digest fallback",
```
tianon, line 196:
```suggestion
			name: "digest fallback",
```

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+				"": "example@sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e",
+			},
+			wantLookup: map[ociregistry.Digest]registry.Reference{
+				"": {
+					Host:       "docker.io",
+					Repository: "library/example",
+					Digest:     "sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e",
+				},
+			},
+			wantDigest: getDigest(""),
```
tianon, line 207: Oh, this one is surprising -- the current implementation handles it correctly, so probably a `TODO` for the future that this ought to return the digest directly (and only have the empty string fallback for tag-lookup) :+1:

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+			wantErr:    "",
+		},
+		{
+			name: "Tag fallback",
```
tianon, line 211:
```suggestion
			name: "tag fallback",
```

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,276 @@ package main
...
+				}
+			}
+
+			if x.wantDigest != nil && lookupDigest == nil {
+				t.Errorf("Unexpected last digest.\ngot:\n%s\n\nexpected:\n%s\n", "<nil>", x.wantDigest)
+			} else if x.wantDigest == nil && lookupDigest != nil {
+				t.Errorf("Unexpected last digest.\ngot:\n%s\n\nexpected:\n%s\n", lookupDigest, "<nil>")
+			} else if lookupDigest != nil && x.wantDigest != nil && *lookupDigest != *x.wantDigest {
+				t.Errorf("Unexpected last digest.\ngot:\n%s\n\nexpected:\n%s\n", lookupDigest, x.wantDigest)
+			}
```
tianon, line 266: Since all these cases are explicit anyhow, I think we should go back to what I suggested in https://github.com/docker-library/meta-scripts/pull/39#discussion_r1583831223, except maybe reverse it (explicit `"nil"` instead of `"empty"`) so the string comparison is simpler and we can avoid the pointers (and the `getDigest` function).  Also, I think we can collapse them (and reverse the conditionals to match the message order so there's less cognitive load while reading it):

```suggestion
			if (lookupDigest == nil && x.wantDigest != "nil") || (lookupDigest != nil && x.wantDigest == "nil") || (lookupDigest != nil && *lookupDigest != x.wantDigest) {
				t.Errorf("Unexpected digest.\ngot:\n%v\n\nexpected:\n%v\n", lookupDigest, x.wantDigest)
			}
```

tianon (commented):
A few more minor nits and one slightly larger one :see_no_evil:

cmd/deploy/input_test.go
LaurentGoderre: The `%v` was the part I was missing. It was expanded because I couldn't get `<nil>` to display properly.

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,279 @@ package main
...
+					t.Fatalf("Expected error not returned: %s", x.wantErr)
+				}
+
+				if !strings.Contains(err.Error(), x.wantErr) {
+					t.Fatalf("Expected error doesn't match.\ngot:\n%q,\n\nexpected to contain:\n%q", err, x.wantErr)
+				}
+			}
+
+			if refsDigest != x.wantDigest {
+				t.Errorf("Digest doesn't match\ngot:\n%s\n\nexpected:\n%s\n", refsDigest, x.wantDigest)
```
whalelines, line 115: Please remove trailing `\n`

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,279 @@ package main
...
+				}
+				return
+			}
+
+			// This allows consistent display of <nil>
+			var wantDigest *digest.Digest
+			if x.wantDigest != "nil" {
+				v := digest.Digest(x.wantDigest)
+				wantDigest = &v
+			}
```
tianon, line 264: Is this purely cosmetic, or was there a functional reason for doing all this pointer juggling?  (in general, we want to avoid pointers where reasonable, and this seems like a place where we don't actually need the pointer except for _slightly_ better display of the error case, which IMO isn't something we should optimize for)

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,279 @@ package main
...
+			if (lookupDigest == nil && wantDigest != nil) ||
+				(lookupDigest != nil && wantDigest == nil) ||
+				(lookupDigest != nil && *lookupDigest != *wantDigest) {
+				t.Errorf("Unexpected digest.\ngot:\n%q\n\nexpected:\n%q", lookupDigest, wantDigest)
```
tianon, line 268: Any particular reason for `%q` vs `%v` here?

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,279 @@ package main
...
+			}
+
+			if !maps.Equal(lookups, x.wantLookup) {
+				t.Errorf("Lookups don't match.\ngot \n%s\n\nexpected\n%s", lookups, x.wantLookup)
```
tianon, line 272: Is `%s` vs some permutation of `%v` intentional here?  It seems weird for a `map` to be printed via `%s` :see_no_evil:

(also `got \n` is weird -- is that trailing space intentional?)

cmd/deploy/input_test.go
LaurentGoderre: Yeah, it's to avoid `nil` vs `<nil>` display

cmd/deploy/input_test.go
LaurentGoderre: It's to make it clear when an empty string is expected

cmd/deploy/input_test.go
LaurentGoderre: Maybe it's just me but it would throw me off to have different representation of nil

tianon: With `%q` it's a little stranger (since we'll get `"nil"`), but IMO that's still plenty clear especially if you check the test code to see exactly which conditional failed, especially since `nil` is an otherwise invalid digest.

tianon: In other words, this will exhibit as either `got "something" expected "nil"` or `got "<nil>" expected "something"` and never `got "<nil>" expected "nil"` so I don't think it's ambiguous in the context of any single error returned from the test (since this one error will never give us _both_ `nil` representations at once).

LaurentGoderre: Done

cmd/deploy/input_test.go
LaurentGoderre: Huh! I just learned about `%#v`

cmd/deploy/input_test.go
```diff
@@ -2,9 +2,272 @@ package main
...
+		{
+			name: "digest mistmatch",
+			rawLookup: map[string]string{
+				"sha256:9ef42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d": "example:foo@sha256:00f42f1d602fb423fad935aac1caa0cfdbce1ad7edce64d080a4eb7b13f7cd9d",
+			},
+			wantErr: "digest on lookup ref should either be omitted or match key:",
+		},
+	} {
+		t.Run(x.name, func(t *testing.T) {
+			lookups, lookupDigest, err := normalizeInputLookup(x.rawLookup)
```
tianon, line 245:
```suggestion
			lookup, lookupDigest, err := normalizeInputLookup(x.rawLookup)
```

:see_no_evil:

tianon (approved):
One final nit I didn't notice on the last round, sorry :innocent: :heart:

I appreciate you sticking with it!

cmd/deploy/input_test.go
LaurentGoderre: Done

tianon (approved):

PR #42, opened 2024-04-19
Update `.go-env.sh` for riscv64

tianon:
Comments inline, but the short version is that riscv64 being unsupported in stable distros makes life a little hard here (https://github.com/docker-library/golang/issues/435), so we need to build from source based on Debian Unstable as the shortest path to success.

### DIFF
--- a/.go-env.sh
+++ b/.go-env.sh
@@ -73,12 +73,35 @@ if [ -t 0 ] && [ -t 1 ]; then
 	fi
 fi
 
-go="$(awk '$1 == "go" { print $2; exit }' "$dir/go.mod")"
-if [[ "$go" == *.*.* ]]; then
-	go="${go%.*}" # strip to just X.Y
+if [ -z "${GOLANG_IMAGE:-}" ]; then
+	go="$(awk '$1 == "go" { print $2; exit }' "$dir/go.mod")"
+	if [[ "$go" == *.*.* ]]; then
+		go="${go%.*}" # strip to just X.Y
+	fi
+	GOLANG_IMAGE="golang:$go"
+
+	# handle riscv64 "gracefully" (no golang image yet because no stable distro releases yet)
+	{
+		if ! docker image inspect --format '.' "$GOLANG_IMAGE" &> /dev/null && ! docker pull "$GOLANG_IMAGE"; then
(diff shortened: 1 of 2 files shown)


tianon:
I wanted to do something more complex that would get updated automatically like checking out the `golang` Git repo and building from there, but then we have to walk back in time to the correct Go version, find the correct variant, etc, so it quickly became diminishing returns.

yosifkit (dismissed):

.go-env.sh
```diff
@@ -73,12 +73,34 @@ if [ -t 0 ] && [ -t 1 ]; then
...
+				# a very rough hack to avoid:
+				#  ERROR: failed to solve: failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to load cache key: subdir not supported yet
+				# (we need buildkit/buildx for --build-context, but newer buildkit than our dockerd might have for build-from-git-with-subdir)
+				export BUILDX_BUILDER="bashbrew-$BASHBREW_ARCH"
+			fi
+			(
+				set -x
+				# TODO make this more dynamic, less hard-coded 🙈
+				# https://github.com/docker-library/golang/blob/ea6bbce8c9b13acefed0f5507336be01f0918f97/1.21/bookworm/Dockerfile
+				docker buildx build --load --tag "$GOLANG_IMAGE" --build-context 'buildpack-deps:bookworm-scm=docker-image://buildpack-deps:unstable-scm' 'https://github.com/docker-library/golang.git#ea6bbce8c9b13acefed0f5507336be01f0918f97:1.21/bookworm'
```
yosifkit, line 96: Perhaps this should be using the `$go` version just grabbed above from the `go.mod`?
```suggestion
				docker buildx build --load --tag "$GOLANG_IMAGE" --build-context 'buildpack-deps:bookworm-scm=docker-image://buildpack-deps:unstable-scm' "https://github.com/docker-library/golang.git#ea6bbce8c9b13acefed0f5507336be01f0918f97:$go/bookworm"
```

tianon: I thought about that!  I decided _not_ to specifically because the `--build-context` relies on the exact `Dockerfile`'s `FROM` line, so it felt more correct to leave it explicit.  However, it should probably also explicitly set `GOLANG_IMAGE=golang:1.21` in this case, _and_ I forgot to commit the change I have locally to `go.mod` so I'll fix that up too. :facepalm: :heart:

yosifkit (approved):

PR #43, opened 2024-04-22
Add more "defense in depth" to collecting manifests for deploy

tianon:
This replicates some protective code we have in Go as well *and* that we currently use aggressive ACLs on Docker Hub (in two different places!) to help protect against.

The end-goal here is to increase Tianon's confidence that we're sufficiently covering the use case of this defensive measure such that we could adjust "deploy" to run on *one* architecture/infra (for deploying to all arch-specific namespaces) instead of running it on *every* architecture, which does have rate limit implications but otherwise simplifies a lot of things (see recent changes adding hacks on hacks to get deploy working on Windows and riscv64 for examples).

### DIFF
--- a/deploy.jq
+++ b/deploy.jq
@@ -8,7 +8,12 @@ def tagged_manifests(builds_selector; tags_extractor):
 			$i
 			| tags_extractor
 			| ..|strings # no matter what "tags_extractor" gives us, this will flatten us to a stream of strings
-		] += $i.build.resolved.manifests
+		] += [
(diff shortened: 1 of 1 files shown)


tianon:
(Opening as a draft so I don't forget about it, but as we discussed offline I agree that we shouldn't consider any _potentially_ dramatic changes like this until at least next week.)

whalelines (approved):
LGTM

yosifkit (approved):

PR #44, opened 2024-05-01
Implement more fine-grained mutexes in registryCache

tianon:
In short, this keeps the "data mutex" as-is, but only acquires it when we have some data to shove in (and only for the limited instructions necessary to update the data), and adds a new mutex per-tag or per-digest to prevent concurrent requests for the exact same upstream content (which is the main reason for this cache in the first place, so feels appropriate).

Without this, our current mutex means that any other efforts to parallelize will ultimately bottleneck on our registry cache.

In order to test this effectively, I've added a `--parallel` flag to `lookup` which just hyper-aggressively runs every single lookup in parallel inside a goroutine.

My first test of this was doing a lookup of the first tag of every DOI repository (so ~147 concurrent lookups in total), and I found it was consistently ~1m57s both with and without this change.  Our Hub rate limiter is pegged at ~100/min, which seems consistent with that result.  If I increase that limit to ~200/min, I was able to achieve a small speedup with this change (~43s down to ~30s).

In other words, for this to actually be effective as a speedup against Docker Hub if we implement parallel deploy, for example, we'll *also* have to increase our rate limit (which I think is fairly safe now that we handle 429 by explicitly emptying the limiter).

In order to *really* test this effectively, I took a different approach.  I spun up a local registry (`docker run -dit -p 127.0.0.1:5000:5000 -p [::1]:5000:5000 --name registry registry`), and copied `hello-world:linux` into it 10,000 times (something like `crane cp hello-world:linux localhost:5000/hello && echo localhost:5000/hello:{1..10000} | xargs -rtn1 -P$(nproc) crane cp localhost:5000/hello` -- could also be done with `jq` and `deploy` if you are ambitious 👀).

Then I used `time ./bin/lookup --parallel $(crane ls --full-ref localhost:5000/hello) > /dev/null` to establish some benchmarks.

Without this change, it's pretty consistently in the 15-20s range on my local system.  With this change, it drops down an order of magnitude to be in the 3-6s range.

### DIFF
--- a/cmd/lookup/main.go
+++ b/cmd/lookup/main.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"sync"
 
 	"github.com/docker-library/meta-scripts/registry"
 )
@@ -22,6 +23,16 @@ func main() {
 	)
 
 	args := os.Args[1:]
+
+	var (
+		parallel = false
+		wg       sync.WaitGroup
+	)
+	if len(args) > 0 && args[0] == "--parallel" {
+		args = args[1:]
+		parallel = true
+	}
+
 	for len(args) > 0 {
 		img := args[0]
 		args = args[1:]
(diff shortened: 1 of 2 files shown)


tianon:
Just for giggles, here's the less speedy (but more fun) version of that `xargs crane` using our `deploy`: `jq -nc '{type: "manifest", refs: ["localhost:5000/hello:" + (range(10000) | tostring)], lookup: { "": "hello-world:linux" }}' | ./bin/deploy`

(I don't know if we'd want to implement `deploy` parallelization across refs in the same "push block" such that this would be faster, but I guess it's possible!)

cmd/lookup/main.go
```diff
@@ -35,53 +46,66 @@ func main() {
...
+				panic(err)
 			}
 		}
 
-		e := json.NewEncoder(os.Stdout)
-		e.SetIndent("", "\t")
-		if err := e.Encode(obj); err != nil {
-			panic(err)
+		if parallel {
+			wg.Add(1)
```
whalelines, line 101: Should there be a limit on how many should be done in parallel?

cmd/lookup/main.go
```diff
@@ -35,53 +46,66 @@ func main() {
...
+			go func(opts registry.LookupOptions) {
+				defer wg.Done()
+				// TODO synchronize output so that it still arrives in-order?  maybe the randomness is part of the charm?
+				do(opts)
+			}(opts)
```
whalelines, line 106: for loop variables and closures

registry/cache.go
```diff
@@ -226,18 +302,24 @@ func (rc *registryCache) PushManifest(ctx context.Context, repo string, tag stri
...
-	// TODO this does *not* need to lock the entire cache during the upstream push (but it *would* be good to block pushing to this specific digest)
-	rc.mu.Lock()
-	defer rc.mu.Unlock()
+	digest := desc.Digest
+	digestKey := cacheKeyDigest(repo, digest)
 
-	// TODO if desc.Size <= manifestSizeLimit, we should technically wrap up the Reader we're given and cache the result so we can shove it directly into the cache, but we currently don't read back blobs we pushed in (and I don't think that's a common use case), so I'm taking the simpler answer of just using this event as a cache bust intead
+	refMu := rc.refMutex(digestKey)
+	refMu.Lock()
+	defer refMu.Unlock()
```
whalelines, line 310: These usages are very similar. Is it possible to refactor to reduce duplication?

whalelines (commented):
I'm guessing since these changes to not change externally observable behavior, the existing test cases cover these changes?

cmd/lookup/main.go
tianon: If this were meant for more than testing extreme concurrency, yes, probably, but the purpose of the new flag on this internal tool is specifically to test massive parallelism (as in my description, I had to get to 10k concurrent requests before I could actually observe a performance difference).

Against Docker Hub, we're already limited by the rate limiter, which will be a more effective limit than anything here anyhow.

cmd/lookup/main.go
tianon: Yeah... making sure we copy things appropriately so we don't leak across invocations.

registry/cache.go
tianon: The best I could do is reduce three lines to two (by having `refMutex` immediately invoke `.Lock()`), which IMO makes it less obvious that we've used the mutex correctly (because you would then have to trace into the function call to see the call to `Lock` to verify that every `Lock` has a matching `Unlock`).

whalelines (approved):
LGTM

yosifkit (approved):

registry/cache.go
```diff
@@ -252,16 +334,22 @@ func (rc *registryCache) PushBlob(ctx context.Context, repo string, desc ociregi
 }
 
 func (rc *registryCache) MountBlob(ctx context.Context, fromRepo, toRepo string, digest ociregistry.Digest) (ociregistry.Descriptor, error) {
-	// TODO this does *not* need to lock the entire cache during the upstream push (but it *would* be good to block pushing to this specific digest)
-	rc.mu.Lock()
-	defer rc.mu.Unlock()
+	// TODO technically we should also be able to safely imply that "fromRepo" has digest here too (assuming MountBlob success), but need to double check whether the contract of the MountBlob API in OCI is such that it's legal for it to return success if "toRepo" already has "digest" (even if "fromRepo" doesn't)
```
tianon, line 337: At least for Docker Hub, we sure could assume this!

```console
$ crane blob tianon/test@sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e > /dev/null
$ crane blob tianon/scratch@sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e > /dev/null
Error: fetching blob tianon/scratch@sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e: GET https://index.docker.io/v2/tianon/scratch/blobs/sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e: BLOB_UNKNOWN: blob unknown to registry; sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e
$ curl -H "$(crane auth token --header --push --mount tianon/scratch tianon/test)" 'https://registry-1.docker.io/v2/tianon/test/blobs/uploads/?mount=sha256:25be82253336f0b8c4347bc4ecbbcdc85d0e0f118ccf8dc2e119c0a47a0a486e&from=tianon/scratch'
{"errors":[{"code":"BLOB_UPLOAD_UNKNOWN","message":"blob upload unknown to registry","detail":"blob upload unknown"}]}
```

(but whether it's _generally_ safe to assume is not guaranteed by the OCI specification, and in fact the opposite might actually be true -- `from` is allowed to be optional :smile:  https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#mounting-a-blob-from-another-repository)

PR #45, opened 2024-05-02
Make "tags" arch-specific again

tianon:
This fixes edge case bugs where tags which share a sourceId but across disparate architectures will be grouped incorrectly (see also https://github.com/docker-library/meta-scripts/pull/27, which is where `tags` was moved to be global instead of arch-specific because I had forgotten about and didn't have an explicit test case for the edge cases where it _does_ matter, which this PR also corrects).
    
This can be seen concretely in https://oci.dag.dev/?image=infosiftr/moby:s390x@sha256:9a423230716c0f934e2fffa46585df1b2babc96ea0a1dd01bbd6ce29c6233aea (where `infosiftr/moby:s390x` should be _just_ the s390x build, but instead gets _all_ the architecture builds which share that same sourceId).

This is currently split into separate concrete commits to ease review, but I don't feel strongly about keeping them separated.

(this PR is dedicated to @paultag, to whom I was explaining how multi-architecture image storage in the registry works, and reached for this image to illustrate, and thus immediately noticed this bug)


tianon:
(this will also require fixing https://github.com/docker-library/meta/blob/90a982d640e236cc31167419152923ee9560830a/.github/workflows/build.yml#L110 again, post-merge)

.test/library/infosiftr-moby
```diff
@@ -0,0 +1,54 @@
...
+Tags: s390x
+GitCommit: 37f1b58c0dc59e6589a0f55849f73d6ef0988cd8
+Directory: infosiftr-moby
+Architectures: s390x
+
+Tags: riscv64
+GitCommit: 37f1b58c0dc59e6589a0f55849f73d6ef0988cd8
+Directory: infosiftr-moby
+Architectures: riscv64
+riscv64-File: Dockerfile.unstable
```
tianon, line 54: Now that it's public, I can link to https://github.com/tianon/bashbrew-tianon/blob/96f417f832852216e0cc276680897c5d63a1fdeb/infosiftr/moby to show where this file comes from (there wasn't really a great reason to keep this private except that I wasn't sure if I wanted to keep doing it, but it's been many years now :joy:).

tianon:
I updated my personal builds to include this change, and so the actual images on Hub are now fixed (which is why my link above is by-digest to illustrate even after it's fixed).

You can see the effect on the JSON in https://github.com/tianon/bashbrew-tianon-meta/commit/371d468ecbbc0016f1ad789ddd8cd5a91ff146c4, and the fixed example s390x image at https://oci.dag.dev/?image=infosiftr/moby:s390x@sha256:d80803750379566d830418b38d0069809891993ec8779b354aa6915d3e5e36bf

whalelines (approved):
If I understand this correctly, it changes the data structure and therefore would require regenerating the various source-of-truth JSON files. How long does it take to regenerate them?

tianon:
The jobs that regenerate them are ~2.5m each when they're load-bearing.

yosifkit (approved):

tianon:
> (this will also require fixing https://github.com/docker-library/meta/blob/90a982d640e236cc31167419152923ee9560830a/.github/workflows/build.yml#L110 again, post-merge)

https://github.com/docker-library/meta/commit/a268733cbb33283a5db6bbd63bd90f2e93ecaa0f :+1:

PR #46, opened 2024-05-15
Maintain inclusion/exclusion criteria (meta, SBOMs, signing) here

tianon:
- move `subset.txt` canonical source here and invert the matching behavior (especially so that we "fail open" and new images are part of the new system automatically)
- add explicit list of Laurent's manually verified SBOM repositories and use that to determine when to `--sbom` on `docker buildx build`
- remove a bunch of `meta.jq` variance by assuming we'll target a BuildKit which can handle attestations/annotations/OCI output
- implement "signing PoC" conditional here to avoid split logic between this repository and the GHA in docker-library/meta

I have manually verified that the output of `bashbrew list --all --repos | jq -L.scripts -rsR 'include "doi"; rtrimstr("\n") | split("\n") | . - repos_anti_subset | join("\n")'` matches our current `subset.txt` exactly. 👍

(probably best to review with ignored whitespace)

### DIFF
--- a/Jenkinsfile.meta
+++ b/Jenkinsfile.meta
@@ -40,6 +40,17 @@ node {
 			sh '''
 				git submodule update --remote --merge .doi
 				git submodule update --remote --merge .scripts
+
+				# TODO once "repos_anti_subset" in "doi.jq" is empty, we can remove this (and all associated usages of "subset.txt" can just be "--all" or go away completely)
+				# in all the places we need to interact with our "subset" it's a lot easier to have an explicit list of what's included, so we'll continue to generate "subset.txt" until it contains the full set
+				bashbrew list --all --repos | jq -L.scripts -rsR '
+					include "doi";
+					rtrimstr("\n")
+					| split("\n")
+					| . - repos_anti_subset
+					| join("\n")
+				' > subset.txt
+				git add subset.txt
 			'''
 		}
 
--- a/doi.jq
+++ b/doi.jq
@@ -1,0 +1,186 @@
+# our old "subset.txt", but inverted
+# this way, we can "fail open" correctly (new images should be part of meta by default)
+# see "Jenkinsfile.meta" for how/why this turns into "subset.txt"
+def repos_anti_subset:
+	[
+		# as we remove items from this list, we need to be careful that none of their *children* are still in the list
+		# (which is why this is sorted in rough "build order" -- that means we can ~safely "pop" off the bottom)
+
+		"almalinux", # direct children: crate
+		"alpine", # direct children: amazoncorretto amazonlinux api-firewall arangodb archlinux bash bonita caddy chronograf consul docker eclipse-mosquitto eclipse-temurin eggdrop erlang fluentd golang haproxy haxe httpd influxdb irssi jobber julia kapacitor kong liquibase memcached nats nats-streaming nginx node notary php postgres python rabbitmq rakudo-star redis registry ruby rust spiped teamspeak telegraf traefik varnish vault znc
+		"alt",
+		"amazonlinux", # direct children: amazoncorretto swift
+		"api-firewall",
+		"arangodb",
+		"archlinux",
+		"bonita",
+		"centos", # direct children: eclipse-temurin ibm-semeru-runtimes percona swift
+		"clearlinux",
+		"clefos",
+		"consul",
+		"crate",
+		"debian", # direct children: adminer aerospike buildpack-deps chronograf clojure couchdb dart emqx erlang haproxy haskell hitch httpd influxdb irssi julia maven memcached mono mysql neo4j neurodebian nginx node odoo openjdk perl php postgres pypy python r-base redis rethinkdb rocket.chat ruby rust spiped swipl unit varnish
+		"eclipse-mosquitto",
+		"eggdrop",
+		"emqx",
+		"express-gateway",
+		"hitch",
+		"jobber",
+		"mageia",
+		"mono",
+		"nats",
+		"nats-streaming",
+		"oraclelinux", # direct children: mysql openjdk percona
+		"percona",
+		"photon",
+		"php", # direct children: backdrop composer drupal friendica joomla matomo mediawiki monica nextcloud phpmyadmin postfixadmin unit wordpress yourls
+		"php-zendserver",
+		"phpmyadmin",
+		"postfixadmin",
+		"r-base",
+		"rethinkdb",
+		"rocket.chat",
+		"rockylinux",
+		"sl",
+		"spiped",
+		"teamspeak",
+		"traefik",
+		"ubuntu", # direct children: buildpack-deps couchbase eclipse-temurin elasticsearch gazebo gradle ibmjava ibm-semeru-runtimes kibana kong logstash mariadb mongo neurodebian odoo php-zendserver rabbitmq ros sapmachine silverpeas swift
+		"varnish",
+		"vault",
+		"yourls",
+		"znc",
+		"adminer",
+		"aerospike",
+		"amazoncorretto", # direct children: jetty maven tomcat
+		"backdrop",
+		"buildpack-deps", # direct children: erlang gcc golang haskell haxe influxdb kapacitor node openjdk perl pypy python rakudo-star ruby rust telegraf
+		"composer", # direct children: drupal
+		"couchbase",
+		"dart",
+		"eclipse-temurin", # direct children: cassandra clojure flink gradle groovy jetty jruby lightstreamer liquibase maven neo4j orientdb solr sonarqube spark storm tomcat tomee unit zookeeper
+		"erlang", # direct children: elixir
+		"flink",
+		"friendica",
+		"gazebo",
+		"gradle",
+		"groovy",
+		"haskell",
+		"haxe",
+		"ibm-semeru-runtimes", # direct children: maven open-liberty tomee websphere-liberty
+		"ibmjava", # direct children: maven websphere-liberty
+		"jetty", # direct children: geonetwork
+		"joomla",
+		"jruby",
+		"kong",
+		"lightstreamer",
+		"liquibase",
+		"matomo",
+		"mediawiki",
+		"monica",
+		"neurodebian",
+		"nextcloud",
+		"node", # direct children: express-gateway ghost mongo-express unit
+		"odoo",
+		"open-liberty",
+		"orientdb",
+		"python", # direct children: hylang plone satosa unit
+		"rakudo-star",
+		"ros",
+		"sapmachine", # direct children: maven
+		"satosa",
+		"silverpeas",
+		"spark",
+		"storm",
+		"swift",
+		"tomcat", # direct children: convertigo geonetwork xwiki
+		"tomee",
+		"websphere-liberty",
+		"zookeeper",
+		"clojure",
+		"convertigo",
+		"geonetwork",
+		"maven",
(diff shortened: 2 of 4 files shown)


doi.jq
```diff
@@ -0,0 +1,186 @@
...
+# input: "build" object (with "buildId" top level key)
+# output: boolean
+def build_should_sign:
+	.build.arch == "amd64" and (
+		.source.arches[.build.arch].tags
+		| map(split(":")[0])
+		| unique
+		| index("notary")
+	)
+;
```
tianon, line 186: Expected usage of this (https://github.com/docker-library/meta/blob/174af2e0f74540b250f365d4cacfe5fce4b45990/.github/workflows/build.yml#L110) would look something like:

```patch
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index fceac66..0a2dd7d 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,7 +107,7 @@ jobs:
           mkdir build
 
           # TODO signing prototype -- starting very small
-          shouldSign="$(jq <<<"$json" '.build.arch == "amd64" and any(.source.arches[.build.arch].tags[]; startswith("notary:"))')"
+          shouldSign="$(jq <<<"$json" -L.scripts 'include "doi"; build_should_sign')"
           [ "$shouldSign" = 'true' ] || [ "$shouldSign" = 'false' ] || exit 1
           echo "shouldSign=$shouldSign" >> "$GITHUB_ENV"
```

meta.jq
```diff
@@ -151,63 +133,50 @@ def build_command:
...
-					if $supportsAnnotationsAndAttestsations then
-						"--provenance=mode=max",
+					"--provenance=mode=max",
+					if build_should_sbom then
 						# see "bashbrew remote arches docker/scout-sbom-indexer:1" (we need the SBOM scanner to be runnable on the host architecture)
 						# bashbrew remote arches --json docker/scout-sbom-indexer:1 | jq '.arches | keys_unsorted' -c
 						if .build.arch as $arch | ["amd64","arm32v5","arm32v7","arm64v8","i386","ppc64le","riscv64","s390x"] | index($arch) then
 							# TODO this needs to be based on the *host* architecture, not the *target* architecture (amd64 vs i386)
 							"--sbom=generator=\"$BASHBREW_BUILDKIT_SBOM_GENERATOR\""
-							# TODO this should also be totally optional -- for example, Tianon doesn't want SBOMs on his personal images
```
tianon, line 171: This design resolves this `TODO` nicely because I'll just patch out `doi.jq` in favor of a `tianon.jq` that always returns `false` for my personal builds :+1:

Jenkinsfile.meta
```diff
@@ -40,6 +40,17 @@ node {
...
+				# TODO once "repos_anti_subset" in "doi.jq" is empty, we can remove this (and all associated usages of "subset.txt" can just be "--all" or go away completely)
+				# in all the places we need to interact with our "subset" it's a lot easier to have an explicit list of what's included, so we'll continue to generate "subset.txt" until it contains the full set
+				bashbrew list --all --repos | jq -L.scripts -rsR '
+					include "doi";
+					rtrimstr("\n")
+					| split("\n")
+					| . - repos_anti_subset
+					| join("\n")
+				' > subset.txt
+				git add subset.txt
```
yosifkit, line 53: So, now that this will generate `subset.txt` from the data in `doi.jq`, maybe we include some or all of the repos in https://github.com/docker-library/meta/pull/31, https://github.com/docker-library/meta/pull/34, https://github.com/docker-library/meta/pull/35, https://github.com/docker-library/meta/pull/36 as part of the changes here (and add a "closes")? That way we can exercise this bit in the first run after merge and also not put those on hold or forget if we decide to merge any of them before this is merged.

meta.jq
```diff
@@ -151,63 +133,50 @@ def build_command:
...
 					@sh "SOURCE_DATE_EPOCH=\(.source.entry.SOURCE_DATE_EPOCH)",
 					# TODO EXPERIMENTAL_BUILDKIT_SOURCE_POLICY=<(jq ...)
 					"docker buildx build --progress=plain",
-					if $supportsAnnotationsAndAttestsations then
-						"--provenance=mode=max",
+					"--provenance=mode=max",
+					if build_should_sbom then
 						# see "bashbrew remote arches docker/scout-sbom-indexer:1" (we need the SBOM scanner to be runnable on the host architecture)
 						# bashbrew remote arches --json docker/scout-sbom-indexer:1 | jq '.arches | keys_unsorted' -c
 						if .build.arch as $arch | ["amd64","arm32v5","arm32v7","arm64v8","i386","ppc64le","riscv64","s390x"] | index($arch) then
```
LaurentGoderre, line 146: Maybe we should move this to the function `build_should_sbom` it woukld make it easier in the future to add sbom to other arch

tianon: I thought about that, but it felt wrong given that the function is entirely based on whether the image/build itself should have an SBOM and _this_ check is whether or not the architecture we're building is even supported by our SBOM-generating image (which is a slightly different concern).  This check is also fundamentally wrong to some extent because it needs to be based on the _host_/build system architecture, _not_ the architecture of the image (AFAIK, BuildKit correctly does _not_ apply `--platform` to the SBOM generator image).

Ultimately, I think it'd be better to refactor the way we set `BASHBREW_BUILDKIT_SBOM_GENERATOR` in the script that invokes this such that it can more appropriately make the decision as to whether we should/do support the SBOM generator image it's pointing at on the current architecture, and then this conditional instead becomes `if env.BASHBREW_BUILDKIT_SBOM_GENERATOR then ...`.

However, with the refactoring in this PR, we can _also_ more easily choose to add SBOMs only for specific architectures of a given repository (or even specific variants if we needed/wanted to!), but I think that's an entirely separate concern to the intent of this `if` statement. :+1:

Jenkinsfile.meta
tianon: Updated :+1:

meta.jq
LaurentGoderre: Ok, sounds good

doi.jq
```diff
@@ -0,0 +1,186 @@
...
+		"geonetwork",
+		"maven",
+		"plone",
+
+		empty
+	]
+;
+
+# a helper for "build_should_sbom"
+def _sbom_subset:
```
yosifkit, line 106: Do we need a check to ensure that something in `_sbom_subset` is not in `repos_anti_subset`?

tianon: :shrug:

I thought about it, but didn't worry about it too much because ideally, we work to shrink `repos_anti_subset` to zero over time, and they are somewhat unrelated concerns (it could be valid for us to have a case of something we've validated the SBOM for, but aren't ready to move yet).

tianon: In other words, I'd love for this to be the source-of-truth for @LaurentGoderre's validation work, and _not_ block that on our ability to add new things. :eyes:

yosifkit (approved):

doi.jq
yosifkit: Oh, good point. 👍

whalelines, line 106: Why do we precede this `def` with an underscore and not `repos_anti_subset`?

doi.jq
```diff
@@ -0,0 +1,186 @@
...
+		"rethinkdb",
+		"ruby",
+		"rust",
+		"solr",
+		"sonarqube",
+		"telegraf",
+		"tomcat",
+		"wordpress",
+		"xwiki",
+		"zookeeper",
```
whalelines, line 158: This list has 50 elements. The tracking spreadsheet has 45. @LaurentGoderre can we explain the difference?

doi.jq
```diff
@@ -0,0 +1,186 @@
...
+;
+
+# input: "build" object (with "buildId" top level key)
+# output: boolean
+def build_should_sign:
+	.build.arch == "amd64" and (
+		.source.arches[.build.arch].tags
+		| map(split(":")[0])
+		| unique
+		| index("notary")
```
whalelines, line 184: Should the images to sign also be extracted into an array like `repos_anti_subset` and `_sbom_subset`?

doi.jq
```diff
@@ -0,0 +1,186 @@
...
+		empty
+	]
+;
+
+# input: "build" object (with "buildId" top level key)
+# output: boolean
+def build_should_sbom:
+	.source.arches[.build.arch].tags
+	| map(split(":")[0])
+	| unique
```
whalelines, line 168: Will this ever produce more than one value? In other words, how could a single DOI repository yield more than one repository portion of the tag?

doi.jq
```diff
@@ -0,0 +1,186 @@
...
+	| _sbom_subset as $subset
+	| any(.[];
+		. as $i
+		| $subset
+		| index($i)
+	)
```
whalelines, line 174: Seems like there should be an easier way to do this

whalelines (approved):
LGTM, just a couple questions

doi.jq
LaurentGoderre: The meta column of the list is at 50. Some images were migrated but don't have SBOM yet or don't have accurate ones (earlier ones that we merged before we haver the current process).

doi.jq
tianon: This one is not _intended_ for external use, but is defined as a function so that it can be used for ad-hoc things if necessary (where `repos_anti_subset` is an intentional interface we use externally elsewhere).

doi.jq
tianon: I decided not to go to that level until we decide on the full plan -- if you look a few lines up, this _also_ limits to `amd64`, which I don't think is what we want for the final solution.  (So for now, this is just porting the existing test to `meta-scripts` as-is.)

doi.jq
tianon: In theory, we could have two separate builds which reference the exact same bytes of build context with the exact same parent chain, in which case they would be joined together under a single `sourceId` (and even share `buildId`s in that case).  In practice, you're absolutely right that this will pretty much always end with an array of length one.

doi.jq
tianon: You would think so, right??  The thing that really gets me is that `contains` is always some weird substring matching, so I have to remember that `index()` is what I want, and it's almost _always_ this reversed usage where I want to check whether a thing I have is in a list of other things for which I have to flip the arguments. :upside_down_face:

It's more often that I'm doing something like https://github.com/docker-library/php/blob/396ead877c1751e756f484e01ac72c93925dfaa8/Dockerfile-linux.template#L321 and thus already have `env.version` but in this case we're iterating.

doi.jq
whalelines: But this is the sbom list, not meta list, right?

tianon: I stole this list from the spreadsheet to begin with, and then added to it the entries from the open PRs :see_no_evil:

IIRC, it was 45 from the sheet and then 6 open PRs which must've had some overlap -- maybe the sheet wasn't updated with all the new open PRs?

tianon: (mostly an easy fixup PR if I've mucked it up, though! :smile:)

doi.jq
whalelines: The list looks correct. I was thinking the spreadsheet might need updating.

tianon:
Updated GHA in https://github.com/docker-library/meta/commit/4158c412b501095179427f51bd4131645a52d457

PR #47, opened 2024-05-22
Migrate more image to the new build and enabled SBOM

LaurentGoderre:


tianon (approved):

tianon:
Might be worth waiting for the Alpine builds/rebuilds to settle a bit before we merge this to make sure we don't get any weird interactions between the older build definitions. :+1:

whalelines:
Has the Alpine backlog resolved? Should this wait until the seccomp issues get resolved?

tianon:
We're very close: https://github.com/docker-library/official-images/issues/16830#issuecomment-2136252379 (our first indication that I finally got all the loose ends)

I agree that we should wait to see that we're fully clear there but I think that's probably the only real blocker to this. :+1:

tianon:
:rocket:

PR #50, opened 2024-05-31
As a temporary balm to our deploy woes, increase the frequency to 2hr

tianon:


tianon:
We should do more of the transition enabled by https://github.com/docker-library/meta-scripts/pull/43, but this should help in the meantime (because that change requires slightly more infrastructure coordination/munging).

tianon:
Gah, this syntax is wrong -- this is the opposite of what I want (two builds per hour instead of one build every two hours).  What I want is not technically supported!

tianon:
To do what I want, I'd have to choose 12 per day, which would possibly use up the whole limit early.

PR #51, opened 2024-06-03
Generate SBOM for the oci-builder

LaurentGoderre:
Add ability to generate SBOM for oci-import builder images

### DIFF
--- a/.test/meta-commands/out.sh
+++ b/.test/meta-commands/out.sh
@@ -137,6 +137,42 @@ jq -s '
 	)
 ' temp/index.json > temp/index.json.new
 mv temp/index.json.new temp/index.json
+# SBOM
+originalImageManifest="$(jq -r '.manifests[0].digest' temp/index.json)"
+SOURCE_DATE_EPOCH=1709081058 \
+	docker buildx build --progress=plain \
+	--load=false \
+	--provenance=false \
+	--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all \
+	--sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR" \
+	--output 'type=oci,tar=false,dest=sbom' \
+	--platform 'linux/amd64' \
+	--build-context "fake=oci-layout://$PWD/temp@$originalImageManifest" \
+	- <<<'FROM fake'
+sbomIndex="$(jq -r '.manifests[0].digest' sbom/index.json)"
+shell="$(jq -r --arg originalImageManifest "$originalImageManifest" '
+	first(
+		.manifests[]
+		| select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
+	) as $attDesc
+	| @sh "sbomManifest=\($attDesc.digest)",
+		@sh "sbomManifestDesc=\(
+			$attDesc
+			| .annotations["vnd.docker.reference.digest"] = $originalImageManifest
+			| tojson
+		)"
+' "sbom/blobs/${sbomIndex/://}")"
+eval "$shell"
+shell="$(jq -r '
+	"copyBlobs=( \([ .config.digest, .layers[].digest | @sh ] | join(" ")) )"
+' "sbom/blobs/${sbomManifest/://}")"
+eval "$shell"
+copyBlobs+=( "$sbomManifest" )
+for blob in "${copyBlobs[@]}"; do
+	cp "sbom/blobs/${blob/://}" "temp/blobs/${blob/://}"
+done
+jq -r --argjson sbomManifestDesc "$sbomManifestDesc" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new
+mv temp/index.json.new temp/index.json
 # </build>
 # <push>
 crane push --index temp 'oisupport/staging-amd64:191402ad0feacf03daf9d52a492207e73ef08b0bd17265043aea13aa27e2bb3f'
--- a/meta.jq
+++ b/meta.jq
@@ -319,9 +319,52 @@ def build_command:
 
 			# TODO consider / check what "crane validate" does and if it would be appropriate here
 
-			# TODO generate SBOM? ... somehow
-
-			empty
+			if build_should_sbom then
+				# we'll trick BuildKit into generating an SBOM for us, then inject it into our OCI layout
+				"# SBOM",
+				"originalImageManifest=\"$(jq -r '.manifests[0].digest' temp/index.json)\"",
+				(
+					[
+						@sh "SOURCE_DATE_EPOCH=\(.source.entry.SOURCE_DATE_EPOCH)",
+						"docker buildx build --progress=plain",
+						"--load=false", "--provenance=false", # explicitly disable a few features we want to avoid
+						"--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all", # disable linting (https://github.com/moby/buildkit/pull/4962)
+						"--sbom=generator=\"$BASHBREW_BUILDKIT_SBOM_GENERATOR\"",
+						"--output 'type=oci,tar=false,dest=sbom'",
+						# TODO also add appropriate "--tag" lines (which would give us a mostly correct "subject" block in the generated SBOM, but we'd then need to replace instances of ${sbomImageManifest#*:} with ${originalImageManifest#*:} for their values to be correct)
+						@sh "--platform \(.source.arches[.build.arch].platformString)",
+						"--build-context \"fake=oci-layout://$PWD/temp@$originalImageManifest\"",
+						"- <<<'FROM fake'", # note: "<<<" is a bashism (so this output must be invoked via bash)
+						empty
+					] | join(" \\\n\t")
+				),
+				"sbomIndex=\"$(jq -r '.manifests[0].digest' sbom/index.json)\"",
+				@sh "shell=\"$(jq -r --arg originalImageManifest \"$originalImageManifest\" \("
+					# https://docs.docker.com/build/attestations/attestation-storage/
+					first(
+						.manifests[]
+						| select(.annotations[\"vnd.docker.reference.type\"] == \"attestation-manifest\")
+					) as $attDesc
+					| @sh \"sbomManifest=\\($attDesc.digest)\",
+						@sh \"sbomManifestDesc=\\(
+							$attDesc
+							| .annotations[\"vnd.docker.reference.digest\"] = $originalImageManifest
+							| tojson
+						)\"
+				" | unindent_and_decomment_jq(4)) \"sbom/blobs/${sbomIndex/://}\")\"",
+				"eval \"$shell\"",
+				@sh "shell=\"$(jq -r \("
+					\"copyBlobs=( \\([ .config.digest, .layers[].digest | @sh ] | join(\" \")) )\"
+				" | unindent_and_decomment_jq(4)) \"sbom/blobs/${sbomManifest/://}\")\"",
+				"eval \"$shell\"",
+				"copyBlobs+=( \"$sbomManifest\" )",
+				"for blob in \"${copyBlobs[@]}\"; do",
+				"\tcp \"sbom/blobs/${blob/://}\" \"temp/blobs/${blob/://}\"",
+				"done",
+				"jq -r --argjson sbomManifestDesc \"$sbomManifestDesc\" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new",
+				"mv temp/index.json.new temp/index.json",
+				empty
+			else empty end
 		] | join("\n")
 	else
 		error("unknown/unimplemented Builder: \($builder)")


LaurentGoderre:
To test, run this within one of the busybox oci folders such as https://github.com/docker-library/busybox/tree/master/latest-1/glibc/amd64

```
docker buildx b . -f ../../../../../builds/meta-scripts/cmd/sbom/Dockerfile -o image
go run ../../../../../builds/meta-scripts/cmd/sbom 4be429a5fbb2e71ae7958bfa558bc637cf3a61baf40a708cb8fff532b39e52d0
```

LaurentGoderre:
The script adds the SBOM with the proper digest filename in the blobs, creates the manifest and adds it to the blobs and finally outputs the manifest to be added to the index

LaurentGoderre:
And to generate the index:

```
set -sex
ocker buildx b . -f ../../../../../builds/meta-scripts/cmd/sbom/Dockerfile -o image; \
go run ../../../../../builds/meta-scripts/cmd/sbom .4be429a5fbb2e71ae7958bfa558bc637cf3a61baf40a708cb8fff532b39e52d0; \
mv index.json input.json; \ 
jq --argjson data "$manifest" '.manifests[.manifests | length] |= . + $data' input.json > index.json  
```

LaurentGoderre:
Ok, here is a pure bash + jq implementation

```sh
docker buildx b . -f ../../../../../builds/meta-scripts/cmd/sbom/Dockerfile -o image;
sbom_digest=$(source="image/image.spdx.json"; cat $source | sha256sum | cut -d ' ' -f1 | xargs -I {} sh -c "mv $source blobs/sha256/{}; echo {}")
sbom_manifest_digest=$( \
    source="sbom_manifest_digest"; \
    jq -cn --arg digest "sha256:$sbom_digest" --arg size $(stat -c %s blobs/sha256/$sbom_digest) -r '{
        "schemaVersion": 2,
        "config": {},
        "layers": [
            {
                "mediaType": "application/vnd.in-toto+json",
                "size": $size,
                "digest": $digest,
                "annotations": {
                    "in-toto.io/predicate-type": "https://spdx.dev/Document"
                }
            }
        ]
    }' > $source; cat $source | sha256sum | cut -d ' ' -f1 | xargs -I {} sh -c "mv $source blobs/sha256/{}; echo {}" \
)
cat <<< $(\
    jq --arg digest "sha256:$sbom_manifest_digest" --arg size $(stat -c %s blobs/sha256/$sbom_manifest_digest) '
        .manifests[0].digest as $ref | 
        {
            "size": $size,
            "digest": $digest,
            "platform": {
                "architecture": "unknown",
                "os": "unknown"
            },
            "annotations": {
                "vnd.docker.reference.digest": $ref,
                "vnd.docker.reference.type": "attestation-manifest"
            }
        } as $manifest |
        .manifests[.manifests | length] |= . + $manifest
    ' index.json \
) > index.json
```

LaurentGoderre:
or full portable...

```bash
docker buildx b . -f - -o image <<EOF
    ARG source=rootfs.tar.gz

    FROM busybox:1.36.1 AS scanner

    RUN --mount=type=bind,target=/scanner,from=docker/scout-sbom-indexer:1-doi \
        --mount=type=bind,target=/image,source=\$source \
        set -ex; \
        mkdir /out; \
        export DOCKER_SCOUT_ENVIRONMENT=indexer; \
        export BUILDKIT_SCAN_DESTINATION=/out; \
        export BUILDKIT_SCAN_SOURCE='/image'; \
        /scanner/sbom-indexer

    FROM scratch

    COPY --from=scanner /out /
EOF
sbom_digest=$(source="image/image.spdx.json"; cat $source | sha256sum | cut -d ' ' -f1 | xargs -I {} sh -c "mv $source blobs/sha256/{}; echo {}")
sbom_manifest_digest=$( \
    source="sbom_manifest_digest"; \
    jq -cn --arg digest "sha256:$sbom_digest" --arg size $(stat -c %s blobs/sha256/$sbom_digest) -r '{
        "schemaVersion": 2,
        "config": {},
        "layers": [
            {
                "mediaType": "application/vnd.in-toto+json",
                "size": $size,
                "digest": $digest,
                "annotations": {
                    "in-toto.io/predicate-type": "https://spdx.dev/Document"
                }
            }
        ]
    }' > $source; cat $source | sha256sum | cut -d ' ' -f1 | xargs -I {} sh -c "mv $source blobs/sha256/{}; echo {}" \
)
cat <<< $(\
    jq --arg digest "sha256:$sbom_manifest_digest" --arg size $(stat -c %s blobs/sha256/$sbom_manifest_digest) '
        .manifests[0].digest as $ref | 
        {
            "size": $size,
            "digest": $digest,
            "platform": {
                "architecture": "unknown",
                "os": "unknown"
            },
            "annotations": {
                "vnd.docker.reference.digest": $ref,
                "vnd.docker.reference.type": "attestation-manifest"
            }
        } as $manifest |
        .manifests[.manifests | length] |= . + $manifest
    ' index.json \
) > index.json
```

tianon:
Very neat! :grinning:

A couple notes:

1. definitely a big fan of doing this in pure shell+`jq` vs Go, _especially_ for similar reasons to one of the notes on https://github.com/docker-library/meta-scripts/pull/53:
   > *Additionally*, we have had a number of issues related to making sure our Go code works on all our architectures, which is less necessary if we make this change (and avoid writing more custom Go in the future that needs to run on *all* workers).

2. I'm very concerned about how much of your `Dockerfile` is effectively reimplementing BuildKit behavior, and in such a way that we then rely on BuildKit to execute it anyhow -- couldn't we do something clever with a simple `Dockerfile` like `FROM oci-import`, and then `--build-context oci-import=oci-layout://temp@sha256:xxx` to convince BuildKit to do the generation more directly?  especially since we have to munge the object/layout afterwards anyhow, we could use that to have BuildKit generate an SBOM for us (which then has the major plus that it references the correct layer IDs!) which we then pluck out of that second OCI layout and inject into ours -- at most I think we'd have to fix up the `subject` field, which we'd need to create using this approach anyhow so that seems pretty reasonable IMO (and could even be just a find/replace instead of a `jq` so we don't have to care about the specific format, ie, scrape the new image index for the image manifest digest, get the attestation object, scrape each object and find/replace the _new_ image manifest digest with the correct older one from the original image, re-digest, inject in our old content, profit)

LaurentGoderre:
Ok, so:

```
docker buildx build --progress=plain --builder=bashbrew-amd64 --provenance=false --build-context "base:latest=oci-layout://$PWD@$(jq -r '.manifests[0].digest' index.json)" --sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR" --output 'type=oci,tar=false,dest=temp/' - <<<'FROM base'
jq '.manifests[0].digest | sub("sha256:";"")' temp/index.json | xargs -I {} mv temp/blobs/sha256/{} temp/index.json
```

LaurentGoderre:
In action: https://explore.ggcr.dev/?image=laurentgoderre689/oci-inmport:busybox-musl-amd64

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,29 @@ jq -s '
 	)
 ' temp/index.json > temp/index.json.new
 mv temp/index.json.new temp/index.json
+SOURCE_DATE_EPOCH=1709081058 \
+	docker buildx build --progress=plain \
+	--provenance=false \
+	--build-context "base:latest=oci-layout://$PWD/temp/@$(jq -r '.manifests[0].digest' index.json)" \
+	--sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR" \
+	--output 'type=oci,tar=false,dest=temp_sbom/' \
```
tianon, line 145: Niiiice, I always forget about `tar=false` -- we should (separately) apply this to the other build so we don't have to extract. :smile:

It's 100% superficial, but I think we should call this directory just `sbom` instead (in hindsight, `temp` was a strange name for our other directory, but we've got a few more external things to port into here like the signing code before we could reasonably swap that).

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,29 @@ jq -s '
...
+	- <<<'FROM base'
+newIndexDigest="$(jq -r '.manifests[0].digest | sub("sha256:";"")' temp_sbom/index.json)"
```
tianon, line 147: I don't love the amount of sha256 assumptions in all this code (and calling things "digest" that aren't actually an OCI digest, but are just the raw checksum).  I _really_ wish there were more tools like `crane` for more easily querying/manipulating OCI layouts, but for now I think we can get away with a shell-ism like `"blobs/${foo/://}"` to get the path to our blobs given an OCI digest.

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,29 @@ jq -s '
...
+sbomManifestDigest="$(jq -r '.manifests[] | select(.platform.os == "unknown").digest | sub("sha256:";"")' temp_sbom/blobs/sha256/$newIndexDigest)"
```
tianon, line 148: Relying on the `unknown` platform makes me nervous -- I'd rather lean on https://docs.docker.com/build/attestations/attestation-storage/ and check for `.annotations["vnd.docker.reference.type"] == "attestation-manifest"` explicitly (ideally we'd _also_ then verify that `vnd.docker.reference.digest` exists and points to a real thing, and maybe even validate some of our assumptions about this clever method, like the layers being identical to our original image, but due to the lack of nice OCI layout tooling I think that's probably too much for now).

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,29 @@ jq -s '
...
+sbomManifest="$(jq -r --arg digest "sha256:$sbomManifestDigest" '.manifests[] | select(.digest == $digest)' temp_sbom/blobs/sha256/$newIndexDigest)"
+sbomDigest="$(jq -r '.layers[0].digest | sub("sha256:";"")' temp_sbom/blobs/sha256/$sbomManifestDigest)"
+cp temp_sbom/blobs/sha256/$sbomDigest temp/blobs/sha256
+cp temp_sbom/blobs/sha256/$sbomManifestDigest temp/blobs/sha256
```
tianon, line 152: Since the way we're building leads to `"subject": []` (and thus _should_ have no references to our intermediate/incorrect image manifest digest/config), this is fine for now, but we probably need a `TODO` in here somewhere about how in the future we'll need to munge the object so it has a proper `subject` list inside the SBOM object, which means a new checksum and slightly more complicated code here (and hopefully we can write that in such a way that it's not _too_ reliant on details of the v0.1 format, which I think we can, but I don't think we need to go to that level right away).  We can take some solace in the fact that building in this way _does_ get us appropriate layer blobs references in the SBOM. :+1:

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,29 @@ jq -s '
...
+cat <<< $( \
+	jq -r \
+	--argjson manifest "$sbomManifest" \
+	'
+		.manifests[0].digest as $digest |
+		.manifests[.manifests | length] |= . + $manifest |
+		.manifests[1].annotations["vnd.docker.reference.digest"] = $digest
+		' \
+	temp/index.json \
+) > temp/index.json
```
tianon, line 162: Oh, this is very cute.  I've never seen `cat <<<$(do something < foo) > foo` used as a moreutils "sponge" replacement before, that's clever.

Unfortunately, however, I think it's probably a little bit _too_ clever because if that inner `jq` fails, the script will continue, so we're probably better off using the same `index.json.new` method as we do in the lines just above this block.

meta.jq
```diff
@@ -26,6 +26,9 @@ def normalize_ref_to_docker:
 	ltrimstr("docker.io/")
 	| ltrimstr("library/")
 ;
+def digest_to_blob:
+	@sh "sub(\"sha256:\";\"\")"
+;
```
tianon, line 31:
```suggestion
```

meta.jq
```diff
@@ -319,8 +322,50 @@ def build_command:
...
+								] | join("\n\t\t")
+							),
+							"temp/index.json",
+							empty
+						] | join(" \\\n\t")
+					),
+					") > temp/index.json",
+					empty
+				] | join(" \\\n")
+			),
```
tianon, line 368: This should _also_ consider `build_should_sbom` so we don't prematurely include SBOMs we didn't intend to :eyes:

Here's a suggested edit that moves things around/adjusts based on my raw script feedback (I took the approach of getting the end result to look the way I wanted it, then worked backwards to making the script create that):

```bash
# ...
mv temp/index.json.new temp/index.json
# SBOM
originalImageManifest="$(jq -r '.manifests[0].digest' temp/index.json)"
SOURCE_DATE_EPOCH=1709081058 \
	docker buildx build --progress=plain \
	--load=false \
	--provenance=false \
	--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all \
	--sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR" \
	--build-context "fake=oci-layout://$PWD/temp@$originalImageManifest" \
	--platform 'linux/amd64' \
	--output 'type=oci,tar=false,dest=sbom' \
	- <<<'FROM fake'
sbomIndex="$(jq -r '.manifests[0].digest' sbom/index.json)"
shell="$(jq -r --arg originalImageManifest "$originalImageManifest" '
	first(
		.manifests[]
		| select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
	) as $attDesc
	| @sh "sbomManifest=\($attDesc.digest)",
		@sh "sbomManifestDesc=\(
			$attDesc
			| .annotations["vnd.docker.reference.digest"] = $originalImageManifest
			| tojson
		)"
' "sbom/blobs/${sbomIndex/://}")"
eval "$shell"
cp "sbom/blobs/${sbomManifest/://}" "temp/blobs/${sbomManifest/://}"
jq -r --argjson sbomManifestDesc "$sbomManifestDesc" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new
mv temp/index.json.new temp/index.json
# </build>
# ...
```

```suggestion
			if build_should_sbom then
				# we'll trick BuildKit into generating an SBOM for us, then inject it into our OCI layout
				"# SBOM",
				"imageManifest=\"$(jq -r '.manifests[0].digest' temp/index.json)\"",
				(
					[
						@sh "SOURCE_DATE_EPOCH=\(.source.entry.SOURCE_DATE_EPOCH)",
						"docker buildx build --progress=plain",
						"--load=false", "--provenance=false", # explicitly disable a few features we want to avoid
						"--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all", # disable linting (https://github.com/moby/buildkit/pull/4962)
						"--sbom=generator=\"$BASHBREW_BUILDKIT_SBOM_GENERATOR\"",
						"--build-context \"fake=oci-layout://$PWD/temp@$imageManifest\"",
						@sh "--platform \(.source.arches[.build.arch].platformString)",
						"--output 'type=oci,tar=false,dest=sbom'",
						# TODO also add appropriate "--tag" lines (which would give us a mostly correct "subject" block in the generated SBOM, but we'd then need to replace instances of ${sbomImageManifest#*:} with ${imageManifest#*:} for their values to be correct)
						"- <<<'FROM fake'", # note: "<<<" is a bashism (so this output must be invoked via bash)
						empty
					] | join(" \\\n\t")
				),
				"sbomIndex=\"$(jq -r '.manifests[0].digest' sbom/index.json)\"",
				@sh "shell=\"$(jq -r --arg imageManifest \"$imageManifest\" \("
					# https://docs.docker.com/build/attestations/attestation-storage/
					first(
						.manifests[]
						| select(.annotations[\"vnd.docker.reference.type\"] == \"attestation-manifest\")
					) as $attDesc
					| @sh \"sbomManifest=\\($attDesc.digest)\",
						@sh \"sbomManifestDesc=\\(
							$attDesc
							| .annotations[\"vnd.docker.reference.digest\"] = $imageManifest
							| tojson
						)\"
				" | unindent_and_decomment_jq(4)) \"sbom/blobs/${sbomIndex/://}\")\"",
				"eval \"$shell\"",
				"cp \"sbom/blobs/${sbomManifest/://}\" \"temp/blobs/${sbomManifest/://}\"",
				"jq -r --argjson sbomManifestDesc \"$sbomManifestDesc\" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new",
				"mv temp/index.json.new temp/index.json",
				empty
			else empty end,

```

tianon (commented):
Nice, this is what I had in mind :muscle: :grinning:

I've got a few notes about things I'd like to add/adjust/make note of, and at the end an explicit suggested change for reorganizing the code to match my notes, but I think it's all pretty minor. :bow: :heart:

(I started with the output, because that's the part I care about the most, and worked backwards from there since it's much harder to reason about the "jq code that generates shell code that embeds more jq code" and where to escape things correctly otherwise IMO :see_no_evil:)

tianon (commented):
Here's a modified version of my initial suggestion, taking into account the added edge of copying the associated config/layer blobs too:

```bash
# SBOM
originalImageManifest="$(jq -r '.manifests[0].digest' temp/index.json)"
SOURCE_DATE_EPOCH=1709081058 \
	docker buildx build --progress=plain \
	--load=false \
	--provenance=false \
	--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all \
	--sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR" \
	--build-context "fake=oci-layout://$PWD/temp@$originalImageManifest" \
	--platform 'linux/amd64' \
	--output 'type=oci,tar=false,dest=sbom' \
	- <<<'FROM fake'
sbomIndex="$(jq -r '.manifests[0].digest' sbom/index.json)"
shell="$(jq -r --arg originalImageManifest "$originalImageManifest" '
	first(
		.manifests[]
		| select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
	) as $attDesc
	| @sh "sbomManifest=\($attDesc.digest)",
		@sh "sbomManifestDesc=\(
			$attDesc
			| .annotations["vnd.docker.reference.digest"] = $originalImageManifest
			| tojson
		)"
' "sbom/blobs/${sbomIndex/://}")"
eval "$shell"
shell="$(jq -r '
	"copyBlobs=( \([ .config.digest, .layers[].digest | @sh ] | join(" ")) )"
' "sbom/blobs/${sbomManifest/://}")"
eval "$shell"
copyBlobs+=( "$sbomManifest" )
for blobDigest in "${copyBlobs[@]}"; do
	cp "sbom/blobs/${blobDigest/://}" "temp/blobs/${blobDigest/://}"
done
jq -r --argjson sbomManifestDesc "$sbomManifestDesc" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new
mv temp/index.json.new temp/index.json
```

Happy to turn this into a PR against your branch if that would make it easier to digest / follow.

(More specific explanations are in inline comments.)

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,37 @@ jq -s '
...
+	--output 'type=oci,tar=false,dest=sbom/' \
+	- <<<'FROM fake'
+sbomIndex="$(jq -r '.manifests[0].digest' sbom/index.json)"
+process=$(jq -r --arg originalImageManifest $originalImageManifest '
+	.manifests[] |
+	select(.annotations["vnd.docker.reference.type"] == "attestation-manifest") |
+	.annotations["vnd.docker.reference.digest"] = $originalImageManifest |
+	{ move: [.digest], descriptor: . }
+' sbom/blobs/${sbomIndex/://})
+sbomManifestDigest="$(jq -rn --argjson process "$process" '$process.descriptor.digest')"
```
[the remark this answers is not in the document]
LaurentGoderre: With the new code though, I only run through each file once and one instruction uses purely memory content. I'm not sure that extra how in memory is worth this level of complexity. I'm also not convinced that a malicious manifest couldn't inject commands to the shell.

LaurentGoderre:
@tianon ok, here is a version that does the same without eval or shell in jq in shell in jq

```bash
{ IFS= read -r -d $'\0\n' sbomManifestDigest
  IFS= read -r -d $'\0\n' descriptor
} < <(
	jq -r --arg originalImageManifest $originalImageManifest '
		.manifests[] |
		select(.annotations["vnd.docker.reference.type"] == "attestation-manifest") |
		.annotations["vnd.docker.reference.digest"] = $originalImageManifest |
		.digest + "\u0000", . 
	' sbom/blobs/${sbomIndex/://}
)
sbomStatementDigest="$(jq -r '.layers[0].digest' sbom/blobs/${sbomManifestDigest/://})"
copyBlobs=( "$sbomManifestDigest" "$sbomStatementDigest" )
for blob in ${copyBlobs[@]}; do
	cp sbom/blobs/${blob/://} temp/blobs/${blob/://}
done
jq -r --argjson descriptor "$descriptor" '.manifests += [ $descriptor ]' temp/index.json > temp/index.json.new
```

.test/meta-commands/out.sh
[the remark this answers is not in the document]
tianon: > I'm also not convinced that a malicious manifest couldn't inject commands to the shell.

I think maybe you're misunderstanding/not familiar with the `@sh` filter in `jq` :sweat_smile:

It exists specifically for protecting against this case, which is how this command is generated (safely) in the first place:

> `@sh`:
> > The input is escaped suitable for use in a command-line for a POSIX shell.

https://jqlang.github.io/jq/manual/v1.6/#:~:text=%40sh%3A,a%20POSIX%20shell.

```console
$ jq -nc '"echo " + ("some `absolutely` $(evil) shell \n code" | @sh)' -r | bash -Eeuo pipefail -x
+ echo 'some `absolutely` $(evil) shell 
 code'
some `absolutely` $(evil) shell 
 code
```

tianon:
Both `IFS` and null bytes are fragile in my experience dealing with shell scripting -- I'd much rather use `jq`'s built-in (and very reliable so far) `@sh` filter to generate correct/safe code.

.test/meta-commands/out.sh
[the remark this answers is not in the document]
LaurentGoderre: I understand it, I just don't like relying on it that way

tianon (commented):
A few more minor suggestions/fixes that I'll have a PR to your branch for shortly (so they're easier to review/digest; GitHub's "suggestions" interface isn't ideal for visualising the actual change): https://github.com/docker-library/meta-scripts/compare/LaurentGoderre:f98f4b0...infosiftr:1485cc8

FWIW, I've been testing these via something like this (which works because `busybox` is using `index.json` in a way we ingest directly, not one we munge -- `crane pull --format oci` could probably also be used, or `git clone --single-branch` of BusyBox directly):

```console
$ rm -rf temp sbom
$ mkdir temp
$ bashbrew context busybox:latest | tar -xvC temp
$ BUILDX_BUILDER=my-builder-name BASHBREW_BUILDKIT_SBOM_GENERATOR=docker/scout-sbom-indexer:1-doi bash -Eeuo pipefail -x sbom.sh
```

(where `sbom.sh` in that directory contains the verbatim copied lines from `.test/meta-commands/out.sh` between `# SBOM` and `# </build>`)

tianon:
"PR-PR" at https://github.com/LaurentGoderre/meta-scripts/pull/1 (happy to see it squashed here though)

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,42 @@ jq -s '
...
+		| select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
+	) as $attDesc
+	| @sh "sbomManifest=\($attDesc.digest)",
+		@sh "sbomManifestDesc=\(
+			$attDesc
+			| .annotations["vnd.docker.reference.digest"] = $originalImageManifest
+			| tojson
+		)"
+' "sbom/blobs/${sbomIndex/://}")"
+eval "$shell"
```
whalelines, line 165: So this pattern is to consolidate two invocations of `jq` into one?

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,42 @@ jq -s '
...
+shell="$(jq -r --arg originalImageManifest "$originalImageManifest" '
+	first(
+		.manifests[]
+		| select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
+	) as $attDesc
+	| @sh "sbomManifest=\($attDesc.digest)",
+		@sh "sbomManifestDesc=\(
+			$attDesc
+			| .annotations["vnd.docker.reference.digest"] = $originalImageManifest
+			| tojson
```
whalelines, line 162: When I remove `| tojson` this fails, but I don't really understand why. It appears to be some strange interaction between `@sh` and `tojson`

```
jq: error (at <stdin>:1): object ({"mediaType...) can not be escaped for shell
```

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,42 @@ jq -s '
...
+		)"
+' "sbom/blobs/${sbomIndex/://}")"
+eval "$shell"
+shell="$(jq -r '
+	"copyBlobs=( \([ .config.digest, .layers[].digest | @sh ] | join(" ")) )"
+' "sbom/blobs/${sbomManifest/://}")"
+eval "$shell"
```
whalelines, line 169: And this one is to leverage `jq`'s `@sh` shell escaping?

.test/meta-commands/out.sh
```diff
@@ -137,6 +137,42 @@ jq -s '
...
+copyBlobs+=( "$sbomManifest" )
+for blob in "${copyBlobs[@]}"; do
+	cp "sbom/blobs/${blob/://}" "temp/blobs/${blob/://}"
+done
+jq -r --argjson sbomManifestDesc "$sbomManifestDesc" '.manifests += [ $sbomManifestDesc ]' temp/index.json > temp/index.json.new
+mv temp/index.json.new temp/index.json
 # </build>
 # <push>
 crane push --index temp 'oisupport/staging-amd64:191402ad0feacf03daf9d52a492207e73ef08b0bd17265043aea13aa27e2bb3f'
```
whalelines, line 178: What is the meaning of this checksum-like tag? Currently the checksum of [this image in Hub](https://explore.ggcr.dev/?image=oisupport%2Fstaging-amd64%3A191402ad0feacf03daf9d52a492207e73ef08b0bd17265043aea13aa27e2bb3f) is sha256:70a227928672dffb7d24880bad1a705b527fab650f7503c191e48a209c4a0d10

whalelines (commented):
A couple questions

.test/meta-commands/out.sh
LaurentGoderre: Yeah, having jq output two values

.test/meta-commands/out.sh
LaurentGoderre: I think this is coming from the test data so it's possible it doesn't match the actual image

whalelines: But why not just push a tag like `meta-scripts-test`?

.test/meta-commands/out.sh
whalelines: Is that motivated by a performance optimization, a DRY-ing of the `first | select`, both, or something else?

whalelines: I found this https://github.com/docker-library/meta-scripts/pull/51#discussion_r1643255385 resolved comment.

.test/meta-commands/out.sh
tianon: That'd be because `@sh` expects to operate on / escape a _string_ for consumption by a shell parser, so we have to use `tojson`/`@json` to convert our object into a string of JSON text which we can then escape for the shell.

.test/meta-commands/out.sh
tianon: Yes, this is to get the list of all child digests as a bash array of properly escaped strings so that we can loop over them without further parsing.

.test/meta-commands/out.sh
tianon: Yes, this is coming from the test data -- this is the "buildId" tags that we use for all our staging pushes.  It looks like a checksum because it _is_ a checksum (of https://github.com/docker-library/meta-scripts/blob/9c6f9d47a24ff6940ee3071e60f2456669f16b23/cmd/builds/main.go#L34-L38).

whalelines (approved):
LGTM

yosifkit (approved):

meta.jq
```diff
@@ -319,9 +319,52 @@ def build_command:
 
 			# TODO consider / check what "crane validate" does and if it would be appropriate here
 
-			# TODO generate SBOM? ... somehow
-
-			empty
+			if build_should_sbom then
```
tianon, line 322: Ahhh, this is currently causing `mips64le` builds to fail because we didn't replicate the architecture check:

https://github.com/docker-library/meta-scripts/blob/f5b18c803010716859916e88cba38834f2b68a7f/meta.jq#L145-L148

Maybe that check should just be part of `doi.jq` in `build_should_sbom` directly instead?

LaurentGoderre: #73

PR #54, opened 2024-06-17
Remove deprecated images

LaurentGoderre:
Follow up to https://github.com/docker-library/official-images/pull/16999


tianon (approved):
Oh nice -- initially I was thinking we should maybe wait until they're fully removed, but this is _just_ removing them from the "anti-subset" and none of them have any direct children, so the end result of this change is that they'll effectively become part of the set of things built in the new system iff they happen to come back with a last-minute update (or with a renewed commitment to maintain within DOI).

PR #55, opened 2024-06-18
Merge new images to the new build

LaurentGoderre:


tianon (approved):
Looks good!  I don't see any reason _not_ to go ahead now :thinking: :+1:

tianon:
(Asking @yosifkit to give it one more set of eyes in case he can think of any reasons I've forgotten before we go ahead with a bunch more :smile: :heart:)

yosifkit (approved):
Nothing I can think of. LGTM

PR #56, opened 2024-06-25
Attempt to speed up deploy

yosifkit:
Now that we have rate limits under control, we should always try a HEAD request on the indexes & manifests before the PUT request (since the PUT is much slower).

Some local testing for `riscv64`, `arm32v6`, and `s390x` using https://github.com/docker-library/meta/commit/6c3188a828643e981b2d1c37cf3851ed9ec2b4b0 shows some significant speed increases. I chose these arches since they were fully pushed and would only do head requests (and my Docker Hub user doesn't have access to push to the arch namespaces anyway).
```console
$ jq -L ../meta-scripts/ 'include "deploy"; arch_tagged_manifests("riscv64") | deploy_objects[]' builds.json > deploy.json
$ time ../meta-scripts/bin/deploy < deploy.json
...
real    0m6.736s
user    0m0.065s
sys     0m0.032s
$ # compared to Jenkins: ~ 2min 49s - 4min 49s

$ jq -L ../meta-scripts/ 'include "deploy"; arch_tagged_manifests("arm32v6") | deploy_objects[]' builds.json > deploy.json
$ time ../meta-scripts/bin/deploy < deploy.json
...
real    0m32.444s
user    0m0.053s
sys     0m0.035s
$ # compared to Jenkins: ~ 5min 16s - 8min 58s

$ jq -L ../meta-scripts/ 'include "deploy"; arch_tagged_manifests("s390x") | deploy_objects[]' builds.json > deploy.json
$ time ../meta-scripts/bin/deploy < deploy.json
...
real    3m12.348s
user    0m0.278s
sys     0m0.252s
$ # compared to Jenkins: ~ 19min 36s - 24min 22s
```

### DIFF
--- a/registry/push.go
+++ b/registry/push.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	// if a manifest or blob is more than this many bytes, we'll do a pre-flight HEAD request to verify whether we need to even bother pushing it before we do so (65535 is the theoretical maximum size of a single TCP packet, although MTU means it's usually closer to 1448 bytes, but this seemed like a sane place to draw a line to where a second request that might fail is worth our time)
+	// if a blob is more than this many bytes, we'll do a pre-flight HEAD request to verify whether we need to even bother pushing it before we do so (65535 is the theoretical maximum size of a single TCP packet, although MTU means it's usually closer to 1448 bytes, but this seemed like a sane place to draw a line to where a second request that might fail is worth our time)
 	BlobSizeWorthHEAD = int64(65535)
 )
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

whalelines (commented):
So this only removes the size check for manifests? Are all our requests still properly accounted for when considering pull and abuse limits?

yosifkit:
> So this only removes the size check for manifests? Are all our requests still properly accounted for when considering pull and abuse limits?

Yup, the rate limiting is active on every request to Docker Hub using this `registry` client (https://github.com/docker-library/meta-scripts/blob/9c6f9d47a24ff6940ee3071e60f2456669f16b23/registry/client.go#L35)

tianon (approved):

tianon:
Because of the way that's implemented, it technically applies to requests to `auth.docker.io` too (any HTTP request that's in the path to our actual lookups gets accounted individually, even including redirects), so it's _technically_ a little tiny bit overaggressive, but I think it's probably better to limit too much than too little. :sweat_smile:

tianon:
Confirmed, `deploy-riscv64` went from ~3m15s down to ~13s :joy:

tianon:
For unrelated reasons, `deploy-arm32v7` has been failing for a little bit, so it had a _definitely_ load-bearing run at ~15m9s with a prior successful run at ~27m, so definitely very net positive. :+1:

PR #59, opened 2024-06-26
Migrate more images

LaurentGoderre:


doi.jq
```diff
@@ -54,10 +53,8 @@ def repos_anti_subset:
 		"mediawiki",
 		"monica",
 		"nextcloud",
-		"python", # direct children: hylang plone satosa unit
```
whalelines, line 57: No SBOM for python?

whalelines (commented):
Question about python

doi.jq
LaurentGoderre: Python was accidentally added to the SBOM list is seems but it wouldn't get them without being on the new build

LaurentGoderre: AAaaah, good point. It never occured to me to do it that way

tianon (approved):

PR #60, opened 2024-06-26
Revert "Revert "Attempt to speed up deploy""

tianon:
This reverts commit 4473698daefb9c5582b904c4df0911232d416443.

A little bit byzantine, but this is #56 re-applied (which was reverted in #58).  See https://github.com/docker-library/meta-scripts/pull/58#issuecomment-2192767052 (and preceding comments) for debugging/justification (and why this should be safe to re-apply).

### DIFF
--- a/registry/push.go
+++ b/registry/push.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

tianon:
The final tally is that our `deploy-riscv64` build is ~20s faster with this change (~3m20s before, ~3m after).

Not a windfall, but definitely measurably better. :heart:

PR #61, opened 2024-06-27
Specify the builder id for provenance

LaurentGoderre:
When generating a provenance using the [GitHub Action Build-Push Action](https://github.com/docker/build-push-action) the action sets the builder id as the URL for the current build. This replicates this behavior.

Example
https://oci.dag.dev/?blob=docker/image-signer-verifier@sha256:56e95d5cbc3afae22c0023f0e386cd04254de1ecd6b98b0b1da483b2a28b1c73&mt=application%2Fvnd.in-toto%2Bjson&size=11949

```json
"builder": {
  "id": "https://github.com/docker/image-signer-verifier/actions/runs/9620065389/attempts/1"
},
```

We are wondering however if the GitHub Action run url is the best value to use of if a different URL would be better suited.

### DIFF
--- a/meta.jq
+++ b/meta.jq
@@ -139,7 +139,7 @@ def build_command:
 					@sh "SOURCE_DATE_EPOCH=\(.source.entry.SOURCE_DATE_EPOCH)",
 					# TODO EXPERIMENTAL_BUILDKIT_SOURCE_POLICY=<(jq ...)
 					"docker buildx build --progress=plain",
-					"--provenance=mode=max",
+					@sh "--provenance=mode=max,builder-id=\(buildkit_provenance_builder_id)",
 					if build_should_sbom then
 						# see "bashbrew remote arches docker/scout-sbom-indexer:1" (we need the SBOM scanner to be runnable on the host architecture)
 						# bashbrew remote arches --json docker/scout-sbom-indexer:1 | jq '.arches | keys_unsorted' -c
--- a/.test/meta-commands/out.sh
+++ b/.test/meta-commands/out.sh
@@ -5,7 +5,7 @@
 # <build>
 SOURCE_DATE_EPOCH=1700741054 \
 	docker buildx build --progress=plain \
-	--provenance=mode=max \
+	--provenance=mode=max,builder-id='https://github.com/docker-library' \
 	--output '"type=oci","dest=temp.tar"' \
 	--annotation 'org.opencontainers.image.source=https://github.com/docker-library/docker.git#6d541d27b5dd12639e5a33a675ebca04d3837d74:24/cli' \
 	--annotation 'org.opencontainers.image.revision=6d541d27b5dd12639e5a33a675ebca04d3837d74' \
--- a/doi.jq
+++ b/doi.jq
@@ -157,6 +157,12 @@ def _sbom_subset:
 	]
 ;
 
+# https://github.com/docker-library/meta-scripts/pull/61 (for lack of better documentation for setting this in buildkit)
+# https://slsa.dev/provenance/v0.2#builder.id
+def buildkit_provenance_builder_id:
+	"https://github.com/docker-library"
+;
+
 # input: "build" object (with "buildId" top level key)
 # output: boolean
 def build_should_sbom:


meta.jq
```diff
@@ -139,7 +139,7 @@ def build_command:
 					@sh "SOURCE_DATE_EPOCH=\(.source.entry.SOURCE_DATE_EPOCH)",
 					# TODO EXPERIMENTAL_BUILDKIT_SOURCE_POLICY=<(jq ...)
 					"docker buildx build --progress=plain",
-					"--provenance=mode=max",
+					"--provenance=\"mode=max,builder-id=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID",
```
jonnystoten, line 142: Missing closing quote here:

```suggestion
					"--provenance=\"mode=max,builder-id=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID\"",
```

jonnystoten (commented):
Thanks @LaurentGoderre ! This is helpful for us when verifying the provenance of DOI because this is a common value to look for across all DOI with provenance attestations.

To fully replicate the docker/build-push-action behavior we'd need to include the "attempt" in the URL too, see [their implementation here](https://github.com/docker/actions-toolkit/blob/main/src/github.ts#L85).

tianon:
> This is helpful for us when verifying the provenance of DOI because this is a common value to look for across all DOI with provenance attestations.

Can you please elaborate more on this?  In what way will it be used/validated?  Just that it starts with `https://github.com/docker-library/meta/actions/...`?

They won't necessarily _all_ look that way -- this PR would also need to account for builds that happen on Jenkins, which does have similar environment variables, but as they won't be terribly stable / long-lived URLs in _either_ case, I'd like to understand more about what the actual underlying use case is before I can reasonably opine on what the best solution/value is.

tianon:
We can close this in favor of https://github.com/docker-library/meta-scripts/pull/76 now, right? :eyes:

LaurentGoderre:
@mrjoelkamp ?

mrjoelkamp:
> We can close this in favor of https://github.com/docker-library/meta-scripts/pull/76 now, right?

We could close this, but there is still some value in keeping it if we are continuing to generate buildkit provenance attestations so that the value is populated with something useful instead of being empty.

It seems reasonable to me to keep this change so long as we are continuing to generate buildkit provenance.

LaurentGoderre:
Ah! @tianon the gha provenance has the right value but the buildkit doesn't. This one populates the buildkit one.

mrjoelkamp:
I must have missed that comment. Because this file is shared with Jenkins, it does make more sense now to leave it empty or hardcode it to some string value (e.g. `buildkit` via `$builder`) but we wouldn't really be using it for anything that I know of.

Having not spent much time reviewing version v0.2 of the SLSA spec, I did check that it requires the [builder.id](https://slsa.dev/spec/v0.2/provenance#builder.id) so it does seem odd/wrong to leave it empty 🤔 

Without over thinking or complicating it, setting `builder.id = "buildkit"` would suffice, thoughts? It's not much better than leaving it `""` but at least it populates a required field 🤷 .

tianon:
Hmm, I see.  I don't think we care _too_ much about this document being "compliant", especially to an officially superseded spec, but setting it to _something_ and not putting too much stock in the particular value is probably fine, yeah.  Since it's supposed to be a URI, maybe https://github.com/docker-library or similar? (so it's still vague, but compliant and accurate to "the entity that executed the invocation")

We should probably follow up with the Docker Docs team to have something about `--provenance=...,builder-id=...` added to https://docs.docker.com/build/metadata/attestations/slsa-provenance/ too :grimacing:
(https://docs.docker.com/build/metadata/attestations/slsa-definitions/ talks about the value and what it means, but not how to set it :see_no_evil:)

So, if we do that instead of a generic "buildkit" (which isn't a URI :sob:), then we need to shift it over to `doi.jq` (because `meta.jq` is, and is intended to be, generic):

```jq
# https://github.com/docker-library/meta-scripts/pull/61 (for lack of better documentation for setting this in buildkit)
# https://slsa.dev/provenance/v0.2#builder.id
def buildkit_provenance_builder_id:
	"https://github.com/docker-library"
;
```

and then something like this in `meta.jq`:

```jq
	@sh "--provenance=\([ "mode=max", "builder-id=\(buildkit_provenance_builder_id)" ] | @csv)",
```

The final raw output of which will be something like `--provenance='"mode=max","builder-id=https://github.com/docker-library"'` -- appropriate shell quoting _and_ CSV quoting, although it would be totally reasonable (and generate slightly nicer looking final shell syntax) to pull `mode=max` out of the CSV quoting, something like this:

```jq
	@sh "--provenance=mode=max,\([ "builder-id=\(buildkit_provenance_builder_id)" ] | @csv)",
```

or even go further and just assume our `builder-id` shouldn't ever contain quotes or a comma, and go all the way to this (with only shell quoting, no CSV quoting):

```jq
	@sh "--provenance=mode=max,builder-id=\(buildkit_provenance_builder_id)",
```

The end result of which would be `--provenance=mode=max,builder-id='https://github.com/docker-library'` (only shell quoting).

tianon:
Lol, I just had to hit "comment" to notice https://docs.docker.com/build/metadata/attestations/slsa-definitions/#:~:text=This%20value%20can%20be%20set%20using,attestation%20parameter. (but it's very much a footnote :sweat_smile:)

> This value can be set using the `builder-id` attestation parameter.

mrjoelkamp:
> Since it's supposed to be a URI, maybe https://github.com/docker-library or similar?

👍 That would be even better! Sounds good to me, and feels more complete than the empty string.

tianon (approved):

PR #63, opened 2024-07-05
Migrate amazoncorretto, ros and gazebo to the new build

LaurentGoderre:


tianon (approved):

tianon:
ROS is one that's notorious for often failing to build come rebuild time, so this might be a noisy one, but I don't think that's a strong reason to not move it (especially since the end goal is "everything").

PR #66, opened 2024-07-15
Migrate more images to the new build

LaurentGoderre:
Depends on https://github.com/docker-library/official-images/pull/17179


yosifkit:
🤔 I'd like to get EOL CentOS dropped from `swift` (https://github.com/docker-library/official-images/pull/17121) before we move it over so that those tags don't get rebuilt. They might just fail to build but I don't want to provide a false sense of "supported" to them.

yosifkit (approved):

LaurentGoderre:
@tianon are you ok with merging this?

tianon (approved):

PR #67, opened 2024-07-29
Removed deprecated images from the list

LaurentGoderre:


doi.jq
```diff
@@ -45,6 +41,12 @@ def repos_anti_subset:
 		"rakudo-star",
 		"silverpeas",
 
+		# deprecated
+		"adminer",
+		"centos", # direct children: eclipse-temurin ibm-semeru-runtimes percona swift
+		"php-zendserver",
+		"sl",
```
whalelines, line 48: The sl bashbrew file has already been removed from docker-library/official-images/library.

whalelines (commented):
Rather than adding the other three here, should we just remove their bashbrew files from docker-library/official-images?

doi.jq
LaurentGoderre: Im more grouping them at the bottom until they can safely be removed

LaurentGoderre:
Im more grouping them at the bottom until they can be safely removed.

tianon:
Agree - we can instead do the two-step and remove them: https://github.com/docker-library/official-images/pull/17276 :eyes:

tianon (approved):

PR #69, opened 2024-08-06
Add timeout to wget

yosifkit:
To hopefully fallback to IPv4 when IPv6 is slow on s390x (at least for the "build" jobs).

Similar to https://github.com/docker-library/oi-janky-groovy/pull/40

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         run: |
           # not doing "uses: docker-library/bashbrew@xxx" because it'll build which is slow and we don't need more than just bashbrew here
           mkdir .bin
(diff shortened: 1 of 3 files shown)


whalelines (approved):
Have we inquired of IBM why IPv6 might be slow on the s390x host they provide?

tianon:
> Have we inquired of IBM why IPv6 might be slow on the s390x host they provide?

No, although that may be prudent to do at some point -- we've seen other networking slowdowns on that instance as well (such as Tailscale being slower than connecting directly).

tianon (approved):
The defaults for `wget` are pretty unreasonable, despite the man page's claims that they ought to be sufficient for anyone (the default is no timeouts for anything except reads, which default to _900 seconds_ before failing).

tianon:
(I don't think we need to call out `s390x` explicitly in the code comment here, but I also don't think it's terribly harmful to do so and certainly not worth another round-trip. :+1:)

PR #70, opened 2024-08-17
Set rewrite-timestamp=true

AkihiroSuda:
Part of:
- docker-library/official-images#16044

This exporter option rewrites the timestamps of the files inside image layers to use $SOURCE_DATE_EPOCH so as to increase reproducibility.

https://github.com/moby/buildkit/blob/v0.15.2/docs/build-repro.md#source_date_epoch

### DIFF
--- a/.test/meta-commands/out.sh
+++ b/.test/meta-commands/out.sh
@@ -6,7 +6,7 @@
 SOURCE_DATE_EPOCH=1700741054 \
 	docker buildx build --progress=plain \
 	--provenance=mode=max,builder-id='https://github.com/docker-library' \
-	--output '"type=oci","dest=temp.tar"' \
+	--output '"type=oci","dest=temp.tar","rewrite-timestamp=true"' \
 	--annotation 'org.opencontainers.image.source=https://github.com/docker-library/docker.git#6d541d27b5dd12639e5a33a675ebca04d3837d74:24/cli' \
 	--annotation 'org.opencontainers.image.revision=6d541d27b5dd12639e5a33a675ebca04d3837d74' \
 	--annotation 'org.opencontainers.image.created=2023-11-23T12:04:14Z' \
--- a/helpers/oci-sbom.sh
+++ b/helpers/oci-sbom.sh
@@ -76,7 +76,7 @@ args=(
 	--load=false --provenance=false # explicitly disable a few features we want to avoid
 	--build-arg BUILDKIT_DOCKERFILE_CHECK=skip=all # disable linting (https://github.com/moby/buildkit/pull/4962)
 	--sbom=generator="$BASHBREW_BUILDKIT_SBOM_GENERATOR"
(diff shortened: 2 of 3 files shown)


tianon:
https://github.com/moby/buildkit/issues/4576#issuecomment-2159501282

> Regarding enabling `rewrite-timestamp=true`, are there any side effects?  In other words, why is the behavior opt-in instead of opt-out or even just enabled by default and/or automatically enabled when an appropriate `SOURCE_DATE_EPOCH` is set?  What are the downsides, and how do we communicate them to _our_ users when they ask us about the metadata of the images we publish?  (Which is a thing that's already surprised quite a few people in our images since we've started setting `SOURCE_DATE_EPOCH` and the timestamps on the metadata of layers of an image were no longer necessarily always linear, which is _technically_ correct, but also surprising behavior, especially after ~10 years of that not being the way this works.)

You replied that it's incompatible with `unpack=true` - should I know what `unpack` is for?  Having to enable this by default if it's generally sane, reasonable, and safe still feels really backwards/wrong and makes me question whether it really _is_ generally sane, reasonable, and safe.

AkihiroSuda:
`unpack=true` means unpacking image blobs as containerd snapshots.
I'll try to implement the support for `unpack=true`, but it does not relate to `docker-library/meta-scripts`.

AkihiroSuda:
ping?

AkihiroSuda:
Rebased

thompson-shaun:
👋 @AkihiroSuda - For this PR specifically what are we seeing as the user benefit? Also, what is the risk to existing users? Are there downstream tools that would be angry?

AkihiroSuda:
> For this PR specifically what are we seeing as the user benefit?

The benefit is the reduction of the diff across builds.
This will help evaluating the potential risk of supply chain attacks.

Note that this `rewrite-timestamp=true` PR only affects the timestamps, and does not affect the file contents.

For reproducing the file contents, the following Dockerfile PRs are still needed:
- https://github.com/docker-library/httpd/pull/250
- https://github.com/tianon/docker-bash/pull/38
- https://github.com/docker-library/haproxy/pull/229
- https://github.com/docker-library/memcached/pull/96
- https://github.com/docker-library/ruby/pull/455
- https://github.com/docker-library/golang/pull/532
- https://github.com/docker-library/wordpress/pull/915

No additional PR is needed for the following repos:
- https://github.com/docker-library/hello-world
- https://github.com/docker-library/docker
- https://github.com/docker-library/busybox
- https://github.com/distribution/distribution-library-image


> Also, what is the risk to existing users?

A build may take an additional couple of minutes for rewriting the timestamps.

> Are there downstream tools that would be angry?

https://github.com/reproducible-containers/diffoci is "angry" by default for timestamp differences, though it has `--ignore-timestamps` flag.

AkihiroSuda:
Rebased

PR #72, opened 2024-09-03
Treat "500 Internal Server Error" and "502 Bad Gateway" the same as a 503

tianon:
We still get spurious errors from our deploy jobs, and almost every time I dig into one, it ends with something like `failed HEAD: 500 Internal Server Error`.

This updates our auto-retry logic to treat 500 (and 502) the same as 503 (ie, up to three retries, subject to our per-request+overall rate limits).

### DIFF
--- a/registry/rate-limits.go
+++ b/registry/rate-limits.go
@@ -24,8 +24,8 @@ func (d *rateLimitedRetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 		// cap request retries at once per second
 		requestRetryLimiter = rate.NewLimiter(rate.Every(time.Second), 1)
 
-		// if we see 3x 503 during retry, we should bail
-		maxTry503 = 3
+		// if we see 3x (503 or 502 or 500) during retry, we should bail
+		maxTry50X = 3
 
 		ctx = req.Context()
 	)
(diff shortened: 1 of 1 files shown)


tianon:
We should maybe _also_ include 502, although we see that a lot less frequently. :thinking:

tianon:
(we also sometimes get `read: connection reset by peer` / unexpected `EOF`, but those are much harder to recover from / auto-retry)

tianon:
> We should maybe _also_ include 502, although we see that a lot less frequently. 🤔

Updated.

tianon:
Well, that's a new failure.  We didn't get that when the automated CI job ran three days ago. :thinking:

```
Error: No files were found with the provided path: .test/.coverage/coverage.*
.test/.coverage/GOCOVERDIR/. No artifacts will be uploaded.
```

tianon:
Oh fun: https://github.com/actions/upload-artifact/releases/tag/v4.4.0

That doesn't explain why it was fine three days ago (when that release is four days ago), but that would explain it!

tianon:
> Oh fun: https://github.com/actions/upload-artifact/releases/tag/v4.4.0
> 
> That doesn't explain why it was fine three days ago (when that release is four days ago), but that would explain it!

Fixed (via https://github.com/docker-library/meta-scripts/commit/f5b18c803010716859916e88cba38834f2b68a7f on main)

registry/rate-limits.go
```diff
@@ -53,9 +53,9 @@ func (d *rateLimitedRetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 			doRetry = true // TODO maximum number of retries? (perhaps a deadline instead?  req.WithContext to inject a deadline?  👀)
 		}
 
-		// 503 should result in a few auto-retries (especially with the automatic retry delay this injects), but up to a limit so we don't contribute to the "thundering herd" too much in a serious outage
-		if res.StatusCode == 503 && maxTry503 > 1 {
-			maxTry503--
+		// certain status codes should result in a few auto-retries (especially with the automatic retry delay this injects), but up to a limit so we don't contribute to the "thundering herd" too much in a serious outage
+		if (res.StatusCode == 503 || res.StatusCode == 502 || res.StatusCode == 500) && maxTry50X > 1 {
```
whalelines, line 57: Why not just `(res.StatusCode >= 500 && res.StatusCode < 600)`?

whalelines (commented):
One question

registry/rate-limits.go
tianon: I thought about it, but hadn't gone through all the possible 5xx to make sure it was sane for us to retry on _all_ of them.  It seems unlikely for Hub to use one of the others unless there's a real error, but I'm sure just because I've put that in writing we'll start seeing it soon. :see_no_evil:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses

whalelines: If it starts responding with 418, we're in trouble

whalelines (approved):
LGTM

yosifkit (approved):

PR #79, opened 2024-09-12
Add OCI functions fort signing

LaurentGoderre:


oci.jq
```diff
@@ -87,3 +87,52 @@ def sort_manifests:
 	sort_by(.platform | sort_split_platform)
 	| sort_attestations
 ;
+
+def index_type:
```
whalelines, line 91: Should this be named something like `index_type_valid`?

whalelines: Or maybe `validate_media_type`?

whalelines (commented):
A question about the naming

oci.jq
```diff
@@ -87,3 +87,52 @@ def sort_manifests:
...
+
+        | .manifests[0] |= (
+            validate_media_type
+            # TODO validate .digest somehow (`crane validate`?) - would also be good to validate all descriptors recursively
+            | if .size < 0 then
+                error("invalid descriptor size: " + .size)
+            else . end
+        )
+    )
+;
```
LaurentGoderre, line 128: This function is this function optimized

https://github.com/docker-library/meta/blob/subset/.github/workflows/build.yml#L191-L241

The new shell function would be:

```sh
validate-oci-layout() {
  local dir="$1"
  jq -L.scripts -s '
    include "oci";
    validate_oci_basic | true
  ' "$dir/oci-layout" "$dir/index.json" || return "$?"
  local manifest
  manifest="$dir/blobs/$(jq -r '.manifests[0].digest | sub(":"; "/")' "$dir/index.json")" || return "$?"
  jq -L.scripts -s '
    include "oci";
    if length != 1 then
      error("unexpected image index document count: " + length)
    else .[0] end
    | validate_index_type

    # TODO more validation?
  ' "$manifest" || return "$?"
}
```

whalelines (approved):
LGTM

oci.jq
```diff
@@ -87,3 +87,52 @@ def sort_manifests:
...
+
+def image_digest($os; $arch):
+  if length != 1 then
+    error("unexpected image index document count: " + (length | tostring))
+  else .[0] end
+  | validate_index_type
+  | .manifests[]
+  | select(.platform.os == $os and .platform.architecture == $arch)
+  | .digest
+;
```
tianon, line 138: What's the use case for this `image_digest` function?  It seems like it's dual purpose "validate the underlying image index" _and_ "extract the (platform specific) image manifest digest" ?

I think in the case of buildkit/buildx output, we can be both more specific _and_ more prescriptive here. :thinking:

ie, I'm thinking we can validate that the index itself contains at least 1 but at most 2 entries, that the second one specifically is our expected provenance object (via annotations / platform), and then we can safely assume the first _must_ be the image (perhaps with some basic media type / size validation). :thinking:

(those are all assumptions we have, but that aren't currently validated anywhere and _should_ be OK if they're broken, but if we're assuming them anyhow, it's valuable to validate them, especially if we're parsing the file anyhow)

tianon (commented):
A few comments/notes - I've tried to leave them in per-function batches so they're hopefully easier to digest / discuss.  I'm certain I've missed some of the detail on the justifications, so please do ask clarifying questions anywhere I've been unclear! :sweat_smile: :heart:

oci.jq
LaurentGoderre: Given the scale of the change for signing, I personally think that would be premature optimization and can easily be refined in a subsequent PR without affecting the function signature.

whalelines: @LaurentGoderre can you expand on how adding more sanity checks would be a premature optimization? After the SBOM incident, it seems prudent to assert on as much as we know.

LaurentGoderre: ok, how about:

```
def image_digest($os; $arch):
    if length != 1 then
        error("unexpected image index document count: \(length)")
    else .[0] end
    | validate_oci_index
    .manifests
    | (
        map(
            select(.annotations["vnd.docker.reference.type"] == "attestation-manifest")
        )
        | if length != 1 then
            error("attestation manifest not found")
        else .[] end
        | .annotations["vnd.docker.reference.digest"]
        | if . == null then
            error("image reference for attestation manifest not found")
        else . end
    ) as $digest
    | map(
        select(.digest == $digest)
    )
    | if length == 0 then
        error("image digest not found in index")
    else .[] end
    | if .platform.architecture != $arch then
        error("architecture does not match expected architecture \($arch)")
    else . end
    | if .platform.os != $os then
        error("os does not match expected os \($os)")
    else . end
    | $digest
;
```

LaurentGoderre: Nope, it's cool how it work but the attestation manifest is not guaranteed to be there (OCI Layout or if the manifest is a referrer)

LaurentGoderre: Maybe I'm not saying it right but I meant in case where the attestations manifest were not in the index but somewhere else (like a referrer)

tianon: Sorry, let me try to rephrase without the added validation component so I can try to be more clear about my concerns (and then we can discuss them better). :bow: :innocent:

My main concerns with this function are the `$os` + `$arch` inputs and the generic name.

The easier of these is the naming; the name ought to be more descriptive of the purpose of the function, and that's easy to fix if we can get a better understanding of how and where it's intended to be used, which it sounds like you've got more use cases in mind than our current use case.  Is that accurate?

My concern with the `$os` and `$arch` inputs is related: it's effectively implementing a tiny subset of https://github.com/containerd/platforms without respect for all the edge cases (for example, `variant`), so as above, I feel like I need to understand better what the intended usage is to provide better/more specific suggestions for how to improve it.

As far as I can see from our current use case, this could be much simpler (and input-less) because the digest we're interested in is _always_ the first one (except in the case of `oci-import` builds, where the `index.json` itself is the "index digest" we're looking for), and if that isn't true, we have an issue and should be failing the build (hence my bringing further validation into the discussion).

Perhaps that helps? :sweat_smile:

LaurentGoderre: But isn't the order also an assumption?

LaurentGoderre: Not needed anymore since we are doing it in bash

LaurentGoderre: we are looking for the digest of the manifest that has the layers. By doing it in bash we can skip the traversing of the hierarchy and find the right one directly.

LaurentGoderre: The bash solution works for oci import just as well as as standard builds.

LaurentGoderre:
@whalelines @tianon the bash method for finding the image digest is in the draft PR for the next step here:
https://github.com/docker-library/meta/pull/41/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R222

yosifkit:
Just one minor bit of feedback is that the indentation should be consistently tabs instead of a mix across the file.

LaurentGoderre:
@yosifkit fixed

tianon:
I had a few comments, but they were almost all minor whitespace suggestions, so I've just applied them in the force push above (https://github.com/docker-library/meta-scripts/compare/7e417acb3a6d5c4577e31b848f0e8471a3312193..a1186bdd958a77bb567a214aa777c5483671647b, https://github.com/docker-library/meta-scripts/compare/7e417acb3a6d5c4577e31b848f0e8471a3312193..a1186bdd958a77bb567a214aa777c5483671647b?w=1).

That also includes some adjustment to `validate_oci_layout_index` so that it outputs the original index again as-is like the other methods (which I think is a property it had before, and a property you've worked to keep in the other methods, so I did it there too even though it's not critical just for consistency).

tianon (approved):

tianon:
It would be nice if we could think of an easy way to test these here without having too many new artifacts committed, but they're a little bit outside of what we can easily test with our existing framework/data, so I'm OK with leaving that out for now. :+1:

tianon:
(oh, I also fixed the "fort" typo that still persisted in the commit message :smile:)

tianon:
Your patience and persistence are appreciated. :+1:

tianon:
> the bash method for finding the image digest is in the draft PR for the next step here: https://github.com/docker-library/meta/pull/41/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R222-R233
> 
> Instead of trying to find the manifest that refers to the image, it looks for the OCI media type of the image and gets the digest via the path.

I'm very concerned by this implementation (effectively scanning a directory and pseudo-parsing JSON via grep/regex), but I guess we'll cross that bridge in a later PR. :see_no_evil:

PR #80, opened 2024-09-13
Modify index munge for verification

LaurentGoderre:
Preserves the image.ref.name annotation and set it to the first arch tag


meta.jq
```diff
@@ -209,11 +209,13 @@ def build_command:
...
 			@sh "jq \("
 				.manifests |= (
-					del(.[].annotations)
-					| unique
+					unique_by([ .digest, .size, .mediaType ])
 					| if length != 1 then
-						error(\"unexpected number of manifests: \" + length)
+						error(\"unexpected number of manifests: \" + (length | tostring))
 					else . end
+					| .[0].annotations += {
```
whalelines, line 216: If you are not deleting the annotations, should this be a `+=` or just an `=`?

whalelines (commented):
One question

meta.jq
LaurentGoderre: We debated this. Both work for the end result. Appending does give a bit of extra data that was otherwise lost.

whalelines: Let me put it another way: why were all the annotations deleted before? Does retaining them cause any issues downstream?

LaurentGoderre: I am hoping for @tianon input on that.' We weren't sure

tianon: That's something that's worth doing some testing with `crane` to make sure it properly ignores the annotations -- it _should_ ignore them, but it is an unknown that's worth validating.

That being said, I'm actually confused why we're setting the annotation explicitly -- switching to that `unique_by` gives us BuildKit's original annotations unmodified, which should be sufficient/correct here, right?

LaurentGoderre: No, the one created by buildkit doesn't work for signing

whalelines: How so? Can you give an example?

LaurentGoderre: Yep. Here is what happens

```
docker buildx b . --platform linux/amd64 \                                                        
-t laurentgoderre689/blah1 \
-t laurentgoderre689/blah2 \
-t laurentgoderre689/blah3 \
-t laurentgoderre689/blah4 \
-t laurentgoderre689/blah5 \
--output '"type=oci","dest=temp.tar"'
```

```json
{
    "schemaVersion": 2,
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "digest": "sha256:930b9b2032c5371cd5ae5b2ffd4d0de0b436606ddf417f7d44f3c73e617daa51",
            "size": 856,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah1:latest",
                "org.opencontainers.image.created": "2024-09-12T19:16:58Z",
                "org.opencontainers.image.ref.name": "latest"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "digest": "sha256:930b9b2032c5371cd5ae5b2ffd4d0de0b436606ddf417f7d44f3c73e617daa51",
            "size": 856,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah2:latest",
                "org.opencontainers.image.created": "2024-09-12T19:16:58Z",
                "org.opencontainers.image.ref.name": "latest"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "digest": "sha256:930b9b2032c5371cd5ae5b2ffd4d0de0b436606ddf417f7d44f3c73e617daa51",
            "size": 856,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah3:latest",
                "org.opencontainers.image.created": "2024-09-12T19:16:58Z",
                "org.opencontainers.image.ref.name": "latest"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "digest": "sha256:930b9b2032c5371cd5ae5b2ffd4d0de0b436606ddf417f7d44f3c73e617daa51",
            "size": 856,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah4:latest",
                "org.opencontainers.image.created": "2024-09-12T19:16:58Z",
                "org.opencontainers.image.ref.name": "latest"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "digest": "sha256:930b9b2032c5371cd5ae5b2ffd4d0de0b436606ddf417f7d44f3c73e617daa51",
            "size": 856,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah5:latest",
                "org.opencontainers.image.created": "2024-09-12T19:16:58Z",
                "org.opencontainers.image.ref.name": "latest"
            }
        }
    ]
}
```

whalelines: Sorry, what about that does not work for signing?

LaurentGoderre: Perhaps a better example

```
docker buildx b . --platform linux/amd64 \
-t laurentgoderre689/blah:blah1 \
-t laurentgoderre689/blah:blah2 \
-t laurentgoderre689/blah:blah3 \
-t laurentgoderre689/blah:blah4 \
-t laurentgoderre689/blah:blah5 \
--output '"type=oci","dest=temp.tar"'
```

```json
{
    "schemaVersion": 2,
    "manifests": [
        {
            "digest": "sha256:e9f9ae2535dc1642338cb63604ee619868d3285177f27cb12ea4abdd14078024",
            "size": 3896,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah:blah1",
                "org.opencontainers.image.created": "2024-09-17T19:58:45Z",
                "org.opencontainers.image.ref.name": "blah1"
            },
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "digest": "sha256:e9f9ae2535dc1642338cb63604ee619868d3285177f27cb12ea4abdd14078024",
            "size": 3896,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah:blah2",
                "org.opencontainers.image.created": "2024-09-17T19:58:45Z",
                "org.opencontainers.image.ref.name": "blah2"
            },
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "digest": "sha256:e9f9ae2535dc1642338cb63604ee619868d3285177f27cb12ea4abdd14078024",
            "size": 3896,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah:blah3",
                "org.opencontainers.image.created": "2024-09-17T19:58:45Z",
                "org.opencontainers.image.ref.name": "blah3"
            },
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "digest": "sha256:e9f9ae2535dc1642338cb63604ee619868d3285177f27cb12ea4abdd14078024",
            "size": 3896,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah:blah4",
                "org.opencontainers.image.created": "2024-09-17T19:58:45Z",
                "org.opencontainers.image.ref.name": "blah4"
            },
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "digest": "sha256:e9f9ae2535dc1642338cb63604ee619868d3285177f27cb12ea4abdd14078024",
            "size": 3896,
            "annotations": {
                "io.containerd.image.name": "docker.io/laurentgoderre689/blah:blah5",
                "org.opencontainers.image.created": "2024-09-17T19:58:45Z",
                "org.opencontainers.image.ref.name": "blah5"
            },
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        }
    ]
}
```

The `"org.opencontainers.image.ref.name"` only refers to the relative tag and not the fully qualified one like `"io.containerd.image.name"` does

whalelines: Thanks, I see it now.

I guess the only outstanding question is whether the including all the annotations cause any issues downstream.

yosifkit: 🤔 if using `org.opencontainers.image.ref.name` requires us to overwrite the value from buildkit and `io.containerd.image.name` already has the correct value, why doesn't the signing just check that annotation instead?

Many versions of buildkit (and other tools) incorrectly (according to [the spec](https://github.com/opencontainers/image-spec/blob/0126d30a90249ba624d134266b662e1d661381f5/annotations.md#L31-L40:~:text=org.opencontainers.image.ref,z0%2D9%5D%2B%0Aseparator%20%3A%3A%3D%20%5B%2D._%3A%40%2B%5D%20%7C%20%22%2D%2D%22)) fill and interpret the `org.opencontainers.image.ref.name` as just the image tag, so rather than trying to force everyone to change how that field is filled/interpreted, we ought to meet them where they are at and just use the `io.containerd.image.name` instead.

LaurentGoderre: We thought about doing that but didn't know how reliable that was and it might be safer to explicitly set the value ourself

mrjoelkamp: another twist is that `docker/build-push-action` and/or `docker/metadata-action` sets `org.opencontainers.image.ref.name` annotation to the full image reference when outputting an OCI layout

mrjoelkamp: I also think that allowing for both `io.containerd.image.name` and/or `org.opencontainers.image.ref.name` for OCI layout verification would be a good idea.  That will require some changes in the verifier code

LaurentGoderre: Fixed here in the meantime

whalelines: Can we dig into these two statements a bit:

1.  if using org.opencontainers.image.ref.name requires us to overwrite the value from buildkit and io.containerd.image.name already has the correct value, why doesn't the signing just check that annotation instead?
2.  `docker/build-push-action` … sets org.opencontainers.image.ref.name annotation to the full image reference when outputting an OCI layout

What is different about the meta build and `docker/build-push-action` that allows BuildKit to set the proper value in the latter but not the former? How is BuildKit getting/setting this value? Should we just ensure the proper value is set in the meta build?

mrjoelkamp: I am partially wrong about `docker/build-push-action` but correct on how we are setting it using `--output ... name=`. I thought the action set this value by default when `--output type=oci` but it looks like we are setting `name=` this manually everywhere `buildkit` is used, including `build-push-action`

mrjoelkamp: [this line ](https://github.com/docker-library/meta-scripts/blob/c703c791d4485b4accdc2f6c119a725919f331c3/.test/meta-commands/out.sh#L9) doesn't set `name=` which is what would allow you to specify the full image reference for the annotation

LaurentGoderre: @mrjoelkamp and I found out that the method from `build-push-action` doesn;t work for our use case because it doesn't work when using containerd.

LaurentGoderre: This is now handled by the signing tool

LaurentGoderre:
@tianon much simpler now!

whalelines (approved):
LGTM

LaurentGoderre:
This index shows what would be pushed using this change: https://explore.ggcr.dev/?image=laurentgoderre689/staging-amd64:d3744c8043a6688aefdff757fc4c6ec66e03ae85c7fa7267d0ee848054cf72cc

PR #82, opened 2024-09-24
Temporarily disable SBOMs on GHA

tianon:
We're chasing an issue with GHA and our SBOM indexer that's leading to *empty* SBOMs being published, which isn't great.  For now, let's disable SBOMs for all our GHA architectures until we figure it out / fix it.

### DIFF
--- a/.test/meta-commands/test.jq
+++ b/.test/meta-commands/test.jq
@@ -1,8 +1,10 @@
 include "meta";
+include "doi"; # TODO remove this
 [
 	first(.[] | select(normalized_builder == "buildkit")),
 	first(.[] | select(normalized_builder == "classic")),
 	first(.[] | select(normalized_builder == "oci-import")),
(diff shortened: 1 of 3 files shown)


.test/meta-commands/test.jq
```diff
@@ -1,8 +1,10 @@
 include "meta";
+include "doi"; # TODO remove this
 [
 	first(.[] | select(normalized_builder == "buildkit")),
 	first(.[] | select(normalized_builder == "classic")),
 	first(.[] | select(normalized_builder == "oci-import")),
+	first(.[] | select(normalized_builder == "oci-import" and build_should_sbom)), # TODO remove this
```
tianon, line 7: This both validates that my conditional is correct and keeps the example of our `oci-import` SBOM code in the validated test output.

yosifkit (approved):

tianon:
Cross-linking the update: https://github.com/docker-library/official-images/pull/17629 :partying_face:

(so we'll revert this soon)

PR #83, opened 2024-09-26
Revert "Temporarily disable SBOMs on GHA"

tianon:
Reverts docker-library/meta-scripts#82 (as of https://github.com/docker-library/official-images/pull/17629, the issue is resolved)


tianon:
(this is a faithful GitHub-generated revert commit, it's pretty vanilla)

tianon:
Well, that's an adorable GitHub quirk -- clicking the "revert" button created a new branch on the origin, which probably should've been blocked by protected branches.  Now even I as an administrator can't delete it because it's protected. :upside_down_face:

PR #85, opened 2024-10-01
Switch from the `subset` branch back to `main`

tianon:
When I created these branches ~a year ago, I had lofty goals that we'd be able to turn on "everything" quickly enough that the `subset` branch would be short-lived and that `main` could stay faithfully tracking the "everything is included" world, but the reality is much messier.

In an attempt to be extremely pragmatic about the reality, my intention is to delete the existing (now extremely outdated) `main` branch and simply rename `subset` to `main`.  It will then organically grow to include "everything" as we've been doing (and eventually we'll even ditch `subset.txt` 🤞).

### DIFF
--- a/.test/provenance/out.json
+++ b/.test/provenance/out.json
@@ -75,7 +75,7 @@
 				"buildType": "https://actions.github.io/buildtypes/workflow/v1",
 				"externalParameters": {
 					"workflow": {
-						"ref": "refs/heads/subset",
+						"ref": "refs/heads/main",
 						"repository": "https://github.com/docker-library/meta",
 						"path": ".github/workflows/build.yml",
 						"digest": {
@@ -98,7 +98,7 @@
 				},
 				"resolvedDependencies": [
 					{
-						"uri": "git+https://github.com/docker-library/meta@refs/heads/subset",
+						"uri": "git+https://github.com/docker-library/meta@refs/heads/main",
 						"digest": {
 							"gitCommit": "0123456789abcdef0123456789abcdef01234567"
 						}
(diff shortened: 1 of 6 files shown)


tianon:
To be clear, this is an explicit Request For Comments -- I don't plan to move on this before next week at the earliest, and will hold as long as anyone has active concerns about doing so.

.test/provenance/out.json
```diff
@@ -107,7 +107,7 @@
 			},
 			"runDetails": {
 				"builder": {
-					"id": "https://github.com/docker-library/meta/.github/workflows/build.yml@refs/heads/subset"
+					"id": "https://github.com/docker-library/meta/.github/workflows/build.yml@refs/heads/main"
```
mrjoelkamp, line 110: This might seem like a wild idea but thoughts on versioning by branch name? It would solve the problem of needing to update `builder.id` if we make a change that warrants a new `builder.id` by incrementing the branch name.

```suggestion
					"id": "https://github.com/docker-library/meta/.github/workflows/build.yml@refs/heads/v1"
```

yosifkit: Would all changes that could affect the `builder.id` be only directly on `meta` though? Could changes here in `meta-scripts` also warrant a `builder.id` change?

mrjoelkamp: Yes, changes in `meta-scripts` could invalidate the SLSA build model that would result in the need to rotate keys and change the `builder.id`. The reason to assign the version from `meta` is that `meta` includes `meta-scripts` as a git submodule and defines the top-level workflow file which is considered the "trusted control plane". 

It seems like a good place to include versioning and we've talked a bit about it in before. It isn't completely necessary but an idea.

tianon: It's a great thought, thanks for raising it!

Given changes here get automatically consumed into `meta` via the automation, I think I'd rather switch to maintaining a new `builder.id` value here in the code when we need to, because that's a lot easier to coordinate correctly than swapping the branch / branch name.

mrjoelkamp: Sound good 😊

tianon:
> Your members will have to manually update their local environments. We'll let them know when they visit the repository, or you can share the following commands.
> 
> ```bash
> git branch -m subset main
> git fetch origin
> git branch -u origin/main main
> git remote set-head origin -a
> ```

PR #88, opened 2024-10-14
Double our Docker Hub rate limit

tianon:
This increases our rate limit from 100/min up to 200/min (with an immediate burst of up to 200 and at most 200 at once).  This should be generally safe since the code will already very aggressively drain the pool of available "tokens" the minute it sees a 429 (thus only doing a trickle of requests anyhow), and we auto-retry on 429, but only after a delay of a full second per request.

If all goes well, this should help mitigate our amd64 deploy jobs that have crept back up to a full hour runtime (without adding further parallelization complexity yet).

### DIFF
--- a/registry/rate-limits.go
+++ b/registry/rate-limits.go
@@ -9,7 +9,7 @@ import (
 
 var (
 	registryRateLimiters = map[string]*rate.Limiter{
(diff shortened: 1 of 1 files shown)


yosifkit (approved):

tianon:
> If all goes well, this should help mitigate our amd64 deploy jobs that have crept back up to a full hour runtime (without adding further parallelization complexity yet).

Confirmed success -- before, `deploy-amd64` was taking almost exactly 1h every hour, and now it's only taking ~29m 52s each time. :+1:

PR #90, opened 2024-10-28
Migrate php and child images to the new build

LaurentGoderre:


doi.jq
```diff
@@ -17,18 +17,7 @@ def repos_anti_subset:
...
-		"joomla",
-		"matomo",
-		"mediawiki",
-		"monica",
-		"nextcloud",
-		"phpmyadmin",
-		"postfixadmin",
-		"yourls",
-
+
```
yosifkit, line 20: image

Github is struggling to represent this diff correctly 😞. There really is a difference between these empty newlines and it looks like this (i.e., the new one has two tabs):
```diff
-
+		
```

Stylistically, I'd say these tabs and the extra empty line on 139 should be removed. Otherwise, LGTM.

LaurentGoderre: Fixed!

tianon: Hilariously, https://github.com/docker-library/meta-scripts/compare/891527fb1ff85b3a85e19cf1028ae5675bf0fde2..05de0308bce4170a64d3f3577637417a65fadd20 _does_ show the tabs correctly (as they're removed). :sob:

tianon (approved):

PR #91, opened 2024-11-04
Implement back off for jobs that are continually failing

yosifkit:
This starts limiting jobs after 24 failures (or GHA triggers) and then limits them to 1 run every 24 job triggers. Since the `trigger-[arch]` jobs runs hourly (triggered from the `meta` job), this equates to a full day of retries and then once a day thereafter.

Also track GHA triggers so they can also have back off when they have triggered many times (i.e. failed to build and push to staging).

Example runs:
 - https://doi-janky.infosiftr.net/job/wip/job/new/job/trigger-gha/10592/
 - https://doi-janky.infosiftr.net/job/wip/job/new/job/trigger-arm32v7/9108/

### DIFF
--- a/.test/jq.sh
+++ b/.test/jq.sh
@@ -13,7 +13,11 @@ for t in "$dir/"*"/test.jq"; do
 	td="$(dirname "$t")"
 	echo -n 'test: '
 	basename "$td"
-	args=( --tab -L "$dir/.." -f "$t" )
+	args=( --tab -L "$dir/.." )
+	if [ -s "$td/in.jq" ]; then
+		jq "${args[@]}" -n -f "$td/in.jq" > "$td/in.json"
+	fi
+	args+=( -f "$t" )
 	if [ -s "$td/in.json" ]; then
 		args+=( "$td/in.json" )
 	else
--- a/.test/meta-queue/test.jq
+++ b/.test/meta-queue/test.jq
@@ -1,0 +1,11 @@
+include "jenkins";
+.pastJobs as $pastJobs
+| .builds
+| get_arch_queue("arm32v7") as $rawQueue
+| $rawQueue | jobs_record($pastJobs) as $newJobs
+| $rawQueue | filter_skips_queue($newJobs) as $filteredQueue
+| (
+	($rawQueue | length) - ($filteredQueue | length)
+) as $skippedCount
+# queue, skips/builds record, number of skipped items
+| $filteredQueue, $newJobs, $skippedCount
--- a/Jenkinsfile.trigger
+++ b/Jenkinsfile.trigger
@@ -13,9 +13,9 @@ env.BASHBREW_ARCH = env.JOB_NAME.minus('/trigger').split('/')[-1] // "windows-am
 def queue = []
 def breakEarly = false // thanks Jenkins...
 
-// this includes the number of attempts per failing buildId
-// { buildId: { "count": 1, ... }, ... }
-def pastFailedJobsJson = '{}'
+// string filled with all images needing build and whether they were skipped this time for recording after queue completion
+// { buildId: { "count": 1, skip: 0, ... }, ... }
+def currentJobsJson = ''
 
 node {
 	stage('Checkout') {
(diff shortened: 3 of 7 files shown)


yosifkit:
Hide whitespace is useful for the later section starting on [new line 224](https://github.com/docker-library/meta-scripts/pull/91/files?w=1#diff-a1eafb3a911bf01d23222ee7808d97d5f227a9608c46843ae4908cdca3408234R224).

tianon:
Nice, this will be really useful for cutting down on the number of builds that we're pretty sure will fail :muscle: :heart:

In the implementation, I don't love how much duplication we have to have, especially those duplicated _and_ reversed conditionals.  Let's spend some time noodling on it to see if we can come up with a clean enough way to merge those together (I'll be working on/thinking about it, but I don't want to discourage other input :smile:). :bow:

Jenkinsfile.trigger
```diff
@@ -17,6 +17,35 @@ def breakEarly = false // thanks Jenkins...
...
+	} else {
+		newFailedJobs[buildId] = [
+			count: 1,
+			firstTime: timeNow,
+			identifier: identifier,
+			lastTime: timeNow,
+			skips: 0,
+			url: url,
+		]
+	}
```
tianon, line 46: If we pull out the "else" block, the common bits of this can collapse (and then we have a more obvious place to "fix" `firstTime` for existing entries later, if we need to:

```suggestion
	newFailedJobs[buildId] = [
		count: 1,
		firstTime: timeNow,
		identifier: identifier,
		lastTime: timeNow,
		skips: 0,
		url: url,
	]
	if (pastFailedJobs[buildId]) {
		// carry forward count (fails/GHA triggers) + time
		newFailedJobs[buildId].count += pastFailedJobs[buildId].count
		newFailedJobs[buildId].firstTime = pastFailedJobs[buildId].firstTime
	}
```

(unfortunately we can't use `Math.min`/`Math.max` to get `firstTime` fixing for "free" because `null` isn't a number, but we could maybe do something clever with the "elvis operator"; `1 ?: 2` => 1, but `null ?: 2` => 2)

tianon: Maybe something like `newFailedJobs[buildId].firstTime = Math.min(pastFailedJobs[buildId].firstTime ?: Long.MAX_VALUE, newFailedJobs[buildId].firstTime)` ? (also totally happy to punt on this - we're not using the value yet, so it's harmless for now and we can keep the easier to read value and we might not even have any "old" data left by the time we want to use it for something)

yosifkit:
Rebased now that #92 is merged. Also, removed the `setFailedJob` function since it can be done in one spot. I'm still considering how to reduce groovy usage and stop it from interfering with the json (why there is the `(long)` cast on times) and how to unify the duplicated `jq` filters.

yosifkit:
Ok, new approach pushed. Grab and use `pastFailedBuilds.json` just for creating current queue. Use `jq` to create and modify the record of "failing"/skipped builds (which now includes all builds triggered since they'll be filtered out on the next successful run). Moved most of the `jq` code to create the queue and job record to small functions in `jenkins.jq`. The should hopefully make it a little easier to test in locally (still needs some automated tests). Also, a bunch of renaming to remove "failed" since it now records all builds.

.test/meta-queue/fake-data.sh
```diff
@@ -0,0 +1,102 @@
...
+		[ null, 1, 0 ], # buildable, tried once
+		[ null, 23, 0 ], # buildable, tried many but less than skip threshold
+		[ null, 24, 0 ], # buildable, tried many, just on skip threshold
+		[ null, 25, 23 ], # buildable, final skip
+		[ null, 25, 24 ], # buildable, no longer skipped
+		[ {}, 3, 0 ], # build "complete" (not queued or skipped)
+		empty # trailing comma
+	] as $data
+	| [ "amd64", "arm32v7", "windows-amd64" ]
+	| to_entries
```
tianon, line 28: Do you still need this?  It seems like both of the following assign `.key` to `$archindex` but then never use it.

jenkins.jq
```diff
@@ -47,3 +49,53 @@ def gha_payload:
...
+	map_values(
+		select(
+			needs_build
+			and .build.arch == $arch
+		)
+		|  if .build.arch | IN("amd64", "i386", "windows-amd64") then
+			# "GHA" architectures (anything we add a "gha_payload" to will be run on GHA in the queue)
+			.gha_payload = (gha_payload | @json)
+		else . end
+		| .identifier = (.source.arches[.build.arch].tags[0] + if .build.arch != $arch then " (\(.build.arch))]" else "" end)
```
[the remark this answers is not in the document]
tianon: On second thought, with the `select` in here, this doesn't even make sense to include anymore :sob: (it will literally never be true and we should think about it when/if we ever bring that behavior back)

```suggestion
		| .identifier = .source.arches[.build.arch].tags[0]
```

.test/meta-queue/fake-data.sh
```diff
@@ -0,0 +1,102 @@
...
+					$skips
+				}
+			}
+		)
+		| add
+	)
+' > "$tmpJson"
+
+head -n1 "$tmpJson" | jq --tab '.' > "$dir/in.json"
+tail -n1 "$tmpJson" | jq --tab '.' > "$dir/past-jobs.json"
```
tianon, line 102: Relevant to my other comment, I did some golfing: :innocent:

```jq
[
	# add new test cases here
	# each item will be used for each architecture generated
	# [ ".build.resloved", "count", "skips" ]
	[ null, 1, 0 ], # buildable, tried once
	[ null, 23, 0 ], # buildable, tried many but less than skip threshold
	[ null, 24, 0 ], # buildable, tried many, just on skip threshold
	[ null, 25, 23 ], # buildable, final skip
	[ null, 25, 24 ], # buildable, no longer skipped
	[ {}, 3, 0 ], # build "complete" (not queued or skipped)
	empty # trailing comma
]
| map(
	("amd64", "arm32v7", "windows-amd64") as $arch
	| ([ $arch, .[] | tostring ] | join("-")) as $buildId
	| {
		# give our inputs cuter names
		resolved: .[0],
		count: .[1],
		skips: .[2],
	}
	| [
		{
			count,
			skips,
		},
		{
			$buildId,
			build: {
				$arch,
				resolved,
				# TODO do we actually *need* more fields here?  or only if we try to "gha_payload" ?
				# (ie, could we swap our "test" to use "arm32v7" instead and avoid more fields here? 👀)
				# (we already have other things testing "gha_payload" so we don't *have* to include that here 😇)
			},
		},
		empty # trailing comma
	]
	| map({ ($buildId): . })
)
| transpose
| map(add)
```

tianon (commented):
A few little things, a few moderate things, and one big one :innocent: :heart:

(Overall this is great though!  Finally doing some backoff is a great plan :bow:)

.test/jq.sh
```diff
@@ -8,8 +8,23 @@ dir="$(readlink -ve "$dir")"
...
+			tests+=( "$file" )
+		else
+			echo >&2 'warning: skipping jq test "'"$t"'", missing test.jq'
+		fi
+	done
+else
+	tests=( "$dir/"*"/test.jq" )
+fi
+
+for t in "${tests[@]}"; do
```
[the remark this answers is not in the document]
yosifkit: I'm not attached to it; it was mostly for convenience of local testing.

.test/meta-queue/fake-data.sh
yosifkit: I mostly like it. Added one small bit at the end so that `in.json` becomes an object instead of a list: `| { pastjobs: .[0], builds: .[1] }`. I also readded `souce.arches.[arch].tags` so that `.identifier` gets filled in.

Force push incoming 🎉

.test/meta-queue/test.jq
```diff
@@ -0,0 +1,11 @@
+include "jenkins";
+.pastjobs as $pastjobs
```
tianon, line 2:
```suggestion
.pastJobs as $pastJobs
```

tianon (approved):

tianon:
@yosifkit noticed after testing this one more time that a queue which is 100% failing but also 100% skipped should be marked as "unstable" in Jenkins (instead of successful), so that's going to be a follow-up

PR #92, opened 2024-11-05
Unify the Trigger build loop for GHA and non-GHA builds

yosifkit:
This is an attempt to unify the queue loop. I wanted to do it without the node allocation and `curl` (i.e., entirely in java/groovy), but the java `URL` class methods are [disallowed](https://github.com/jenkinsci/script-security-plugin/blob/df2fc45f229c75a4ab8c88800bee49370462eb7b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist#L77-L81) in the sandbox (security concern of reading local files of the Jenkins node).

So, without installing the [HTTP Request](https://plugins.jenkins.io/http_request/) plugin, we can't post to a URL without `curl` (and thus a node allocation) and it is unfortunately mostly unmaintained: "This plugin is up for adoption! We are looking for new maintainers"

#91 will be rebased on this if/when this is merged.


Jenkinsfile.trigger
```diff
@@ -126,55 +133,31 @@ node {
...
+						c += pastFailedJobs[buildObj.buildId].count
+					}
+					// TODO maybe implement some amount of backoff? keep first url/endTime?
+					newFailedJobs[buildObj.buildId] = [
+						count: c,
+						identifier: identifier,
+						url: res.absoluteUrl,
+						endTime: (res.startTimeInMillis + res.duration) / 1000.0, // convert to seconds
+					]
+					error()
```
tianon, line 159: Since we got rid of the `echo`, we could stuff `res.result` in here (because otherwise it's an exception without a message, which is kind of silly but we don't really have any more information here than "the job failed" :joy:)

```suggestion
					error(res.result)
```

Jenkinsfile.trigger
```diff
@@ -70,6 +71,7 @@ node {
 									| index($arch)
 								)
 							)
+							| .payload = (gha_payload | @json)
```
tianon, line 74: I think I'd rather call this `gha_payload` explicitly, and _only_ generate/set/return it when it's actually necessary (which we can then use further down as the determiner so that this becomes the only block that knows/cares about `BASHBREW_ARCH=gha`):

```suggestion
							| if env.BASHBREW_ARCH == "gha" then
								.gha_payload = (gha_payload | @json)
							else . end
```

Jenkinsfile.trigger
```diff
@@ -85,36 +87,41 @@ node {
...
+		def json = writeJSON(json: buildObj, returnText: true)
+		echo(json) // for debugging/data purposes
+
+		// "catchError" to set "stageResult" :(
+		catchError(message: 'Build of "' + identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
+			if (env.BASHBREW_ARCH == 'gha') {
+				node {
 					withEnv([
-						'json=' + json,
+						'payload=' + buildObj.payload
```
tianon, line 113:
```suggestion
						'payload=' + buildObj.gha_payload,
```

tianon (changes requested):
Just a few minor things -- shuffling things around a tiny bit but overall I like it :eyes: :+1: :heart:

Jenkinsfile.trigger
```diff
@@ -85,36 +89,41 @@ node {
...
+		def json = writeJSON(json: buildObj, returnText: true)
+		echo(json) // for debugging/data purposes
+
+		// "catchError" to set "stageResult" :(
+		catchError(message: 'Build of "' + identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
+			if (buildObj.gha_payload) {
+				node {
 					withEnv([
-						'json=' + json,
+						'payload=' + buildObj.gha_payload
```
tianon, line 115:
```suggestion
						'payload=' + buildObj.gha_payload,
```

:see_no_evil:

Jenkinsfile.trigger
```diff
@@ -126,55 +133,31 @@ node {
...
+			} else {
+				def res = build(
+					job: 'build-' + env.BASHBREW_ARCH,
+					parameters: [
+						string(name: 'buildId', value: buildObj.buildId),
+					],
+					propagate: false,
+					quietPeriod: 5, // seconds
+				)
+				// TODO do something useful with "res.result" (especially "res.result != 'SUCCESS'")
```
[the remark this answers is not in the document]
tianon: :eyes:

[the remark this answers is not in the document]
yosifkit: Oops, I deleted the wrong `TODO` comment. 🙈

tianon (approved):

PR #96, opened 2024-11-20
Update Jenkinsfile.* to use shallow clones of meta

tianon:
We don't need a full (unshallow) clone for any of these -- the jobs don't look at the commit history and we can make new commits without having history locally.

This will help on architectures that have been struggling to clone the repository since the move (because a "full" clone is now bigger than it was when we started all this), but is also generally better all around.

(I've tested this successfully on `mips64le`'s failing `build` job: https://doi-janky.infosiftr.net/job/meta/job/mips64le/job/build/10430/console)

### DIFF
--- a/Jenkinsfile.build
+++ b/Jenkinsfile.build
@@ -20,6 +20,11 @@ node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 			]],
 			branches: [[name: '*/main']],
 			extensions: [
+				cloneOption(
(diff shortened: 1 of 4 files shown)


yosifkit:
> Hmm, should we go even further and add an explicit refspec and turn on the "honor refspec on clone" setting too so it only clones the single branch we care about? 

I don't think it matters either way since we almost never do branches on the main fork.

yosifkit (approved):

PR #99, opened 2024-11-23
Add ability for parallel Jenkins multi-arch builds

yosifkit:
Trigger all jenkins builds (adding them to jenkins queue) and then parallel wait on them. We control the concurrency by the number of executors per arch (or if we add multiple machines that match the build label).

### DIFF
--- a/Jenkinsfile.build
+++ b/Jenkinsfile.build
@@ -1,15 +1,25 @@
-// one job per arch (for now) that just builds "the top thing" (triggered by the meta-update job)
+// any number of jobs per arch that build the specified buildId (triggered by the respective trigger job)
 properties([
-	disableConcurrentBuilds(),
+	// limited by one job per buildId so that the same build cannot run concurrently
+	throttleJobProperty(
+		limitOneJobWithMatchingParams: true,
+		paramsToUseForLimit: 'buildId',
+		throttleEnabled: true,
+		throttleOption: 'project',
+	),
 	disableResume(),
 	durabilityHint('PERFORMANCE_OPTIMIZED'),
 	parameters([
 		string(name: 'buildId', trim: true),
+		string(name: 'identifier', trim: true, description: '(optional) used to set <code>currentBuild.displayName</code> to a meaningful value earlier'),
 	]),
 ])
 
 env.BASHBREW_ARCH = env.JOB_NAME.minus('/build').split('/')[-1] // "windows-amd64", "arm64v8", etc
 env.BUILD_ID = params.buildId
+if (params.identifier) {
+	currentBuild.displayName = params.identifier + ' (#' + currentBuild.number + ')'
+}
 
 node('multiarch-' + env.BASHBREW_ARCH) { ansiColor('xterm') {
 	stage('Checkout') {
--- a/Jenkinsfile.trigger
+++ b/Jenkinsfile.trigger
@@ -102,57 +102,20 @@ if (breakEarly) { return } // thanks Jenkins...
 
 // new data to be added to the past-jobs.json
 // { lastTime: unixTimestamp, url: "" }
-def buildCompletionData = [:]
+buildCompletionData = [:]
 
-for (buildObj in queue) {
-	stage(buildObj.identifier) {
-		//def json = writeJSON(json: buildObj, returnText: true)
-		//echo(json) // for debugging/data purposes
-
-		// "catchError" to set "stageResult" :(
-		catchError(message: 'Build of "' + buildObj.identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
-			if (buildObj.gha_payload) {
-				node {
-					withEnv([
-						'payload=' + buildObj.gha_payload,
-					]) {
-						withCredentials([
-							string(
-								variable: 'GH_TOKEN',
-								credentialsId: 'github-access-token-docker-library-bot-meta',
-							),
-						]) {
-							sh '''
-								set -u +x
-
-								# https://docs.github.com/en/free-pro-team@latest/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event
-								curl -fL \
-									-X POST \
-									-H 'Accept: application/vnd.github+json' \
-									-H "Authorization: Bearer $GH_TOKEN" \
-									-H 'X-GitHub-Api-Version: 2022-11-28' \
-									https://api.github.com/repos/docker-library/meta/actions/workflows/build.yml/dispatches \
-									-d "$payload"
-							'''
-						}
-					}
-					// record that GHA was triggered (for tracking continued triggers that fail to push an image)
-					buildCompletionData[buildObj.buildId] = [
-						lastTime: System.currentTimeMillis() / 1000, // convert to seconds
-						url: currentBuild.absoluteUrl,
-					]
-				}
-			} else {
-				def res = build(
-					job: 'build',
-					parameters: [
-						string(name: 'buildId', value: buildObj.buildId),
-					],
-					propagate: false,
-					quietPeriod: 5, // seconds
+// list of closures that we can use to wait for the jobs on.
+def waitQueue = [:]
+def waitQueueClosure(identifier, buildId, externalizableId) {
+	return {
+		stage(identifier) {
+			// "catchError" to set "stageResult" :(
+			catchError(message: 'Build of "' + identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
+				def res = waitForBuild(
+					runId: externalizableId,
+					propagateAbort: true, // allow cancelling this job to cancel all the triggered jobs
 				)
-				// record the job failure
-				buildCompletionData[buildObj.buildId] = [
+				buildCompletionData[buildId] = [
 					lastTime: (res.startTimeInMillis + res.duration) / 1000, // convert to seconds
 					url: res.absoluteUrl,
 				]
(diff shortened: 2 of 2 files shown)


Jenkinsfile.build
```diff
@@ -1,15 +1,18 @@
 // one job per arch (for now) that just builds "the top thing" (triggered by the meta-update job)
 properties([
-	disableConcurrentBuilds(),
```
tianon, line 3: Oof, hmm, something to think about here:

https://github.com/docker-library/meta/blob/f595ec2375e11540c4e26352274b05247953d763/.github/workflows/build.yml#L28-L30

I wish Jenkins had something like GitHub's concurrency groups -- this prevents the same `buildId` from even _attempting_ to build twice at the same time, and with this PR the only thing that prevents that is that these are canonically triggered by the "trigger" job, which is fine for the normal case, but when things go wrong and we're running "build" by hand to debug, nothing will stop "trigger" from firing and potentially clobbering the thing we're testing. :thinking:

I think the closest thing Jenkins has is "Lockable Resources" (https://plugins.jenkins.io/lockable-resources/), but they're a really awful experience and IIRC nothing cleans them up, so we can't reasonably put an arbitrary number of those into the system. :sob:

yosifkit: I think the ["Throttle Concurrent Builds" plugin](https://plugins.jenkins.io/throttle-concurrents/) can at least get things with the same parameters from running at the same time: https://www.gusi.me/2022/05/06/Disable-concurrent-builds-based-on-parameters.html. It will still be quasi-queued though.

yosifkit:
```groovy
	throttleJobProperty(
		limitOneJobWithMatchingParams: true,
		paramsToUseForLimit: 'identifier, buildId',
		throttleEnabled: true,
		throttleOption: 'project',
	),
```
Ok, I did a local test using the "Throttle Builds" plugin with this job config and found that it can stop concurrent jobs from adding to the Jenkins queue when specific parameters match already queued/running jobs.

image

If the build job activation comes from something like the trigger job that uses `waitForStart :true`, then the trigger job will be stuck waiting to add it to the queue and not start anything else. This should be fine since trigger is the only one that should be starting them and so would only happen if we manually had started a build.

Jenkinsfile.build
```diff
@@ -1,15 +1,25 @@
...
+		limitOneJobWithMatchingParams: true,
+		paramsToUseForLimit: 'buildId',
+		throttleEnabled: true,
+		throttleOption: 'project',
+	),
 	disableResume(),
 	durabilityHint('PERFORMANCE_OPTIMIZED'),
 	parameters([
+		string(name: 'identifier', trim: true),
 		string(name: 'buildId', trim: true),
```
tianon, line 14: Let's put the "required" parameter first and give the `identifier` a description so we remember what it is while triggering the job manually (also doubles as a "load-bearing" code comment :smile:):

```suggestion
		string(name: 'buildId', trim: true),
		string(name: 'identifier', trim: true, description: '(optional) used to set <code>currentBuild.displayName</code> to a meaningful value earlier'),
```

Jenkinsfile.trigger
```diff
@@ -102,69 +102,98 @@ if (breakEarly) { return } // thanks Jenkins...
...
-def buildCompletionData = [:]
+buildCompletionData = [:]
 
-for (buildObj in queue) {
-	stage(buildObj.identifier) {
-		//def json = writeJSON(json: buildObj, returnText: true)
-		//echo(json) // for debugging/data purposes
+// list of closures that we can use to wait for the jobs on.
+def waitQueue = [:]
+def addToWait(identifier, buildId, externalizableId) {
```
tianon, line 109: I recall your original version of this function actually _was_ adding the given parameters to the queue, but given the complications with that IMO we should rename this to make it more clear it only returns the closure and adding it to the queue is still something the caller is responsible for; maybe something like this?

```suggestion
def waitQueueClosure(identifier, buildId, externalizableId) {
```

Jenkinsfile.trigger
```diff
@@ -102,69 +102,98 @@ if (breakEarly) { return } // thanks Jenkins...
...
+	return {
+		// "catchError" to set "stageResult" :(
+		catchError(message: 'Build of "' + identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
+			stage(identifier) {
```
tianon, line 113: I know you tested and it _should_ work in this order, but I think it's slightly clearer if we swap these two (both so we don't catch `stage()` itself somehow failing and so it's clear our stage happens and then we set the state of it from within not without):

```suggestion
		stage(identifier) {
			// "catchError" to set "stageResult" :(
			catchError(message: 'Build of "' + identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
```

Jenkinsfile.trigger
```diff
@@ -102,69 +102,98 @@ if (breakEarly) { return } // thanks Jenkins...
...
+					error(res.result)
+				}
+			}
+		}
+	}
+}
+
+// stage to wrap up all the build job triggers that get waited on later
+stage('trigger') {
+	for (int i = 0; i < queue.size(); i++) {
```
tianon, line 133: Do you remember what the error is when you left this with the previous syntax so we could note it in a comment?

Jenkinsfile.trigger
```diff
@@ -102,69 +102,98 @@ if (breakEarly) { return } // thanks Jenkins...
...
-					]
 				}
 			} else {
+				// why not parallel these build() invocations?
+				// jenkins parallel closures get started in a randomish order, ruining our sorted queue
 				def res = build(
 					job: 'build',
 					parameters: [
+						string(name: 'identifier', value: buildObj.identifier),
 						string(name: 'buildId', value: buildObj.buildId),
```
tianon, line 178: (same order swap)

```suggestion
						string(name: 'buildId', value: buildObj.buildId),
						string(name: 'identifier', value: buildObj.identifier),
```

Jenkinsfile.trigger
```diff
@@ -102,69 +102,98 @@ if (breakEarly) { return } // thanks Jenkins...
...
 					],
 					propagate: false,
-					quietPeriod: 5, // seconds
+					// trigger these quickly so they all get added to Jenkins queue in "queue" order
+					quietPeriod: 0, // seconds
+					// we'll wait on them after they are all queued
+					waitForStart: true,
```
tianon, line 184: We can't have a `quietPeriod` on these (even though that would be nice) because it messes with `waitForStart` (and we would then be "captive" for the entire quiet period) -- I think we should probably note that here. :thinking:

Maybe something like this?

```suggestion
					// trigger these quickly so they all get added to Jenkins queue in "queue" order (also using "waitForStart" means we have to wait for the entire "quietPeriod" before we get to move on and schedule more)
					quietPeriod: 0, // seconds
					// we'll wait on the builds in parallel after they are all queued (so our sorted order is the queue order)
					waitForStart: true,
```

tianon (commented):
Mostly cosmetics! :smile: :heart:

Jenkinsfile.trigger
yosifkit: I got it from reading https://www.jenkins.io/doc/pipeline/examples/#jobs-in-parallel and https://www.jenkins.io/doc/pipeline/examples/#parallel-from-grep, but it seems like the function is now enough to scope it correctly (tested locally), so I'll switch back to "for-in"

Jenkinsfile.trigger
yosifkit: Seems sane. I should also move it inside the `stage(buildObj.identifier)` on https://github.com/docker-library/meta-scripts/pull/99/files#diff-a1eafb3a911bf01d23222ee7808d97d5f227a9608c46843ae4908cdca3408234R139, but that means I'll need to duplicate it for the `else` block but that seems fine.

Jenkinsfile.trigger
```diff
@@ -102,69 +102,100 @@ if (breakEarly) { return } // thanks Jenkins...
...
-						lastTime: System.currentTimeMillis() / 1000, // convert to seconds
-						url: currentBuild.absoluteUrl,
-					]
 				}
-			} else {
+			}
+		} else {
+			// "catchError" to set "stageResult" :(
+			catchError(message: 'Build of "' + buildObj.identifier + '" failed', buildResult: 'UNSTABLE', stageResult: 'FAILURE') {
+
```
tianon, line 173: This blank line is extraneous, but it doesn't bother me to leave it in. :+1:

```suggestion
```

Jenkinsfile.trigger
```diff
@@ -102,69 +102,100 @@ if (breakEarly) { return } // thanks Jenkins...
...
-				}
+				waitQueue[buildObj.identifier] = waitQueueClosure(buildObj.identifier, buildObj.buildId, res.externalizableId)
 			}
 		}
 	}
 }
 
+// wait on all the 'build' jobs that were queued
+if (waitQueue.size() > 0) {
+	parallel waitQueue
```
tianon, line 196: I usually prefer to be explicit about the `()` in single-argument invocations like this (although I don't usually go as far as "naming" the parameter), but I don't think it really matters much either way. :+1:

```suggestion
	parallel(waitQueue)
```

tianon (approved):
I think this is great :+1:

I do think we should wait to merge it until _after_ Thanksgiving at this point, though :eyes: :turkey:

Jenkinsfile.trigger
```diff
@@ -102,69 +102,100 @@ if (breakEarly) { return } // thanks Jenkins...
...
-				node {
-					withEnv([
-						'payload=' + buildObj.gha_payload,
-					]) {
-						withCredentials([
-							string(
-								variable: 'GH_TOKEN',
-								credentialsId: 'github-access-token-docker-library-bot-meta',
-							),
+buildCompletionData = [:]
```
yosifkit, line 105: > Did you forget the `def` keyword? WorkflowScript seems to be setting a field named buildCompletionData (to a value of type LinkedHashMap) which could lead to memory leaks or other issues.

This is a warning that I saw in Jenkins and just wanted to document our choice here.  We needed global scope (no `def`) so that the `waitQueueClosure` function could access this variable. We chose no `def` over `@Field` because `@Field \n def` is more confusing than a simple global variable.

https://stackoverflow.com/a/50573082

PR #100, opened 2024-12-10
Save firstTime this build was attempted in past-jobs.json

yosifkit:
just some extra data so we can see what time this started failing without having to calculate based on skips, count, and assuming the job ran every hour.

I meant to include this in https://github.com/docker-library/meta-scripts/pull/91, but it got lost in the refactors.

Also drop temporary migration of `pastFailedJobs.json` since it isn't necessary anymore.

### DIFF
--- a/Jenkinsfile.trigger
+++ b/Jenkinsfile.trigger
@@ -52,10 +52,7 @@ node {
 					! wget --timeout=5 -qO past-jobs.json "$JOB_URL/lastSuccessfulBuild/artifact/past-jobs.json" \\
 					|| ! jq 'empty' past-jobs.json \\
 				; then
-					# temporary migration of old data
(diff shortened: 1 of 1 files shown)


Jenkinsfile.trigger
```diff
@@ -211,7 +208,8 @@ node {
 
 					jq <<<"$currentJobsJson" '
 						# merge the two objects recursively, preferring data from "buildCompletionDataJson"
-						. * ( env.buildCompletionDataJson | fromjson )
+						# save firstTime if it isn't set yet
+						map_values(.firstTime //= .lastTime) * ( env.buildCompletionDataJson | fromjson )
```
tianon, line 212: Can't have `'` because we're inside `'` :smile:  also if we split this up the diff is slightly better and it's slightly more obvious that we're adding data _and_ munging data since they're separate

```suggestion
						. * ( env.buildCompletionDataJson | fromjson )
						# save firstTime if it is not set yet
						| map_values(.firstTime //= .lastTime)
```

tianon (approved):

PR #105, opened 2025-01-23
Implement parallelism in deploy

tianon:
This requires the input to be in the correct order (children first), but that was already an implied constraint.

In my testing, this takes our `.test/test.sh --deploy` block testing deploy from ~5s down to ~2.5s.

I used "show timestamps" in our GHA CI jobs to compare them also, and they're currently at ~7s, and after this change they drop to ~1s.

I've tested this with my personal deploy job, which was taking ~2m40s before, and is ~1m18s after.

PR #110, opened 2025-01-31
(re-)Implement parallelism in deploy

tianon:
This requires the input to be in the correct order (children first), but that was already an implied constraint.

In my testing, this takes our `.test/test.sh --deploy` block testing deploy from ~5s down to ~2.5s.

When we originally introduced this, it collided with some other aging infrastructure in a bad way, kicking over everything, and had to be reverted.  We've since changed that and this should be safe to re-introduce.

See also:
- https://github.com/docker-library/meta-scripts/pull/105#issuecomment-2625991599
- https://github.com/docker-library/meta-scripts/pull/107
- https://github.com/docker-library/meta-scripts/pull/106

### DIFF
--- a/cmd/deploy/main.go
+++ b/cmd/deploy/main.go
@@ -18,7 +18,25 @@ func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
 	defer stop()
 
-	// TODO --dry-run ?
+	var (
+		args = os.Args[1:]
+
+		// --parallel
+		parallel bool
+	)
+	for len(args) > 0 {
+		arg := args[0]
+		args = args[1:]
+
+		// TODO --dry-run
+		switch arg {
+		case "--parallel":
+			parallel = true
+
+		default:
+			panic("unknown argument: " + arg)
+		}
+	}
 
 	// TODO the best we can do on whether or not this actually updated tags is "yes, definitely (we had to copy some children)" and "maybe (we didn't have to copy any children)", but we should maybe still output those so we can trigger put-shared based on them (~immediately on "definitely" and with some medium delay on "maybe")
 
@@ -103,7 +121,8 @@ func main() {
 			normal := normal.clone()
 
 			wg.Add(1)
-			go func() {
+			// (making a function instead of direct "go func() ..." so we can support the --parallel toggle)
+			f := func() {
 				defer wg.Done()
 
 				if mutex != nil {
@@ -177,7 +196,12 @@ func main() {
 					logText += "@" + string(desc.Digest)
 				}
 				fmt.Println(successPrefix + logText)
-			}()
+			}
+			if parallel {
+				go f()
+			} else {
+				f()
+			}
 		}
 	}
 
--- a/cmd/deploy/main.go
+++ b/cmd/deploy/main.go
@@ -7,6 +7,11 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"sync"
+
+	"github.com/docker-library/meta-scripts/registry"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 func main() {
@@ -32,6 +37,11 @@ func main() {
 		panic(err)
 	}
 
+	// a set of RWMutex objects for synchronizing the pushing of "child" objects before their parents later in the list of documents
+	// for every RWMutex, it will be *write*-locked during push, and *read*-locked during reading (which means we won't limit the parallelization of multiple parents after a given child is pushed, but we will stop parents from being pushed before their children)
+	childMutexes := sync.Map{}
+	wg := sync.WaitGroup{}
+
 	dec := json.NewDecoder(stdout)
 	for dec.More() {
 		var raw inputRaw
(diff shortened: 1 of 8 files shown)


tianon:
As noted in https://github.com/docker-library/meta-scripts/pull/105#issuecomment-2625991599, if this is something @yosifkit agrees seems relatively safe to re-introduce, we should wait at _least_ until next Monday to do so. :bow:

tianon:
We could also do what we did in `cmd/lookup` and have an explicit `--parallel` boolean flag for this: :thinking:

https://github.com/docker-library/meta-scripts/blob/5ea66801c9c7bb686e270233f286c8a4a2b9990c/cmd/lookup/main.go#L31-L34

yosifkit:
I think an explicit `--parallel` flag would be good. Then we can turn it on or off as we test and verify our proxy without reverting it again.

tianon:
- Implement explicit `--parallel` flag

- Add `--dry-run` flag
    
    This swaps the output to be input-compatible such that we can verify the results of a deploy, especially our new "filtered deploy" (we can verify that the filtering was warranted, but out-of-band instead of in-band).
    
    This also allows us to do more/better testing (see `.test` changes, especially `test.sh`).

cmd/deploy/main.go
```diff
@@ -48,26 +84,173 @@ func main() {
...
+
+				if dryRun {
+					needsDeploy, err := normal.dryRun(ctx, ref)
+					if err != nil {
+						fmt.Fprintf(os.Stderr, "%s -- ERROR: %v\n", failurePrefix+ref.String()+logSuffix, err)
+						panic(err) // TODO exit in a more clean way (we can't use "os.Exit" because that causes *more* errors 😭)
+					}
+					if needsDeploy {
+						normal.Refs = []registry.Reference{ref}
+						j, err := json.MarshalIndent(normal, "", "\t")
```
yosifkit, line 214: You wanted a note to remind you to add a test to ensure the `json` roundtrip produces output that is the same as the input.

tianon: yes, good call -- one of our existing test cases fails because this assumption has a hole in it! :smile:

```
--- FAIL: TestNormalizeInput (0.00s)
    --- FAIL: TestNormalizeInput/copy_manifest_(tag) (0.00s)
        --- FAIL: TestNormalizeInput/copy_manifest_(tag)/roundtrip (0.00s)
            input_test.go:457: normalize error: localhost:5000/example:test: missing data (and lookup is not a single item or fallback with digest or tag)
```

```json
{
	"type": "manifest",
	"refs": [ "localhost:5000/example:test" ],
	"lookup": { "": "tianon/true:oci" }
}
```

normalizes to:

```json
{
	"type": "manifest",
	"refs": [ "localhost:5000/example:test" ],
	"lookup": { "": "tianon/true" },
	"copyFrom": "tianon/true:oci"
}
```

Note especially that when we ignore the `copyFrom` field like `Normalize` does, we no longer know that it's the `oci` tag of `tianon/true` we're trying to copy! :sweat_smile:

I'm not sure what the best way to fix this is. :thinking:

tianon: Got to a pretty clean solution in dfb595a. :+1:

tianon:
comparing https://github.com/docker-library/meta-scripts/actions/runs/13119586724/job/36602129641#logs to https://github.com/docker-library/meta-scripts/actions/runs/13125173171/job/36619923417?pr=110

```diff
--- before.txt	2025-02-03 16:08:00.384383528 -0800
+++ after.txt	2025-02-03 16:06:06.851327211 -0800
@@ -3,10 +3,13 @@
 cmd/builds/main.go:			loadCacheFromFile		76.0%
 cmd/builds/main.go:			saveCacheToFile			78.6%
 cmd/builds/main.go:			main				80.7%
+cmd/deploy/input.go:			clone				100.0%
 cmd/deploy/input.go:			normalizeInputRefs		100.0%
 cmd/deploy/input.go:			normalizeInputLookup		100.0%
-cmd/deploy/input.go:			NormalizeInput			81.3%
-cmd/deploy/main.go:			main				77.1%
+cmd/deploy/input.go:			NormalizeInput			80.3%
+cmd/deploy/input.go:			do				87.5%
+cmd/deploy/input.go:			dryRun				78.6%
+cmd/deploy/main.go:			main				91.0%
 cmd/lookup/main.go:			main				74.6%
 om/om.go:				Keys				100.0%
 om/om.go:				Get				100.0%
@@ -21,7 +24,7 @@
 registry/cache.go:			GetBlob				100.0%
 registry/cache.go:			GetManifest			100.0%
 registry/cache.go:			GetTag				91.2%
-registry/cache.go:			resolveBlob			83.3%
+registry/cache.go:			resolveBlob			100.0%
 registry/cache.go:			ResolveManifest			100.0%
 registry/cache.go:			ResolveBlob			100.0%
 registry/cache.go:			ResolveTag			84.6%
@@ -31,7 +34,8 @@
 registry/client.go:			Client				57.1%
 registry/client.go:			EntryForRegistry		86.7%
 registry/lookup.go:			Lookup				87.5%
-registry/push.go:			EnsureManifest			78.9%
+registry/manifest-children.go:	ParseManifestChildren		100.0%
+registry/push.go:			EnsureManifest			73.2%
 registry/push.go:			CopyManifest			71.4%
 registry/push.go:			EnsureBlob			77.8%
 registry/push.go:			CopyBlob			65.5%
@@ -47,4 +51,4 @@
 registry/synthesize-index.go:	setRefAnnotation		100.0%
 registry/synthesize-index.go:	normalizeManifestPlatform	83.9%
 registry/user-agent.go:		RoundTrip			83.3%
-total:					(statements)			80.1%
+total:					(statements)			81.5%
```

yosifkit (approved):

PR #114, opened 2025-02-21
Save full list of entries for every source

tianon:
When we normalize sources by deduplicating on `sourceId`, we end up losing a little bit of data that's important for being able to reconnect `source` objects with their lines in `library/foo`.

This resolves that by adjusting our `entry` object to be a vector of `entries` instead, but sorted using the same `SOURCE_DATE_EPOCH` tiebreaker we used previously so that `.entries[0]` is the same as our old `.entry`, but we keep the full list of values for later use/lookup/cross-referencing.

I have additionally verified that `meta.jq` here was the only place we are actively consuming from the `.entry` object (currently), so this should be fully sufficient (no changes necessary elsewhere).

### DIFF
--- a/.test/builds.json
+++ b/.test/builds.json
@@ -77,15 +77,17 @@
     "source": {
       "sourceId": "76d7d7d66aeb62eb797c8475407e6cd2b6ad262957a622035f81fb93e532b36b",
       "reproducibleGitChecksum": "52d3d61b1e9d12310646ec3935698a5825bf568cecc5b5fbbf4e4c4a59e53e8b",
-      "entry": {
-        "GitRepo": "https://github.com/docker-library/docker.git",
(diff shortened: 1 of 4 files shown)


tianon:
(I dedicate this change to @mrjoelkamp :saluting_face: :sob: :heart:)

tianon:
An (easy) alternative we could do would be to calculate this `.entries` value, and then at the very very end, convert to `.entry` + `.otherEntries` or something, so the churn is smaller. :thinking:

tianon:
(but that makes using it for the intended purpose more annoying because the data is in two places, so I think this change is probably more correct/useful as-is and the one-time heavy indentation change is worth it)

yosifkit (approved):

PR #116, opened 2025-02-27
Run meta job from git triggers

yosifkit:
The current git commit to image on `library/` build and publish pipeline looks like this:

```console
meta -> trigger -> build -> meta -> deploy -> put-shared
```

With meta needing to reconcile `souces.json` and `builds.json` so that builds can be triggered and so that deploy can push from staging to arch-specific namespaces, if we run it more often, we can improve the "time to pullable" from `library` on Docker Hub.

Still a TODO: skipping behavior since this will no longer be guaranteed hourly triggers (and often could be more than hourly)

### DIFF
--- a/Jenkinsfile.meta
+++ b/Jenkinsfile.meta
@@ -3,13 +3,14 @@ properties([
 	disableResume(),
 	durabilityHint('PERFORMANCE_OPTIMIZED'),
 	pipelineTriggers([
-		cron('@hourly'),
+		githubPush(),
+		cron('@daily'), // check periodically, just in case
 	]),
 ])
 
 node {
 	stage('Checkout') {
-		checkout(scmGit(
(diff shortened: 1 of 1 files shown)


tianon (approved):

Jenkinsfile.meta
```diff
@@ -3,13 +3,14 @@ properties([
 	disableResume(),
 	durabilityHint('PERFORMANCE_OPTIMIZED'),
 	pipelineTriggers([
-		cron('@hourly'),
+		githubPush(),
+		cron('@daily'), // check periodically, just in case
```
tianon, line 7: We have to move this back to `@hourly` for now so we can pick up completed builds (until we figure out / implement a better way to have those triggers make their way back into the system).

PR #118, opened 2025-02-28
Speed up sources.sh even more

yosifkit:
by improving slow `jq` section in sources.sh (building upon the gains from https://github.com/docker-library/meta-scripts/pull/115)

Local speed improvement:
```console
$ cp sources.json sources-copy.json
$ # before change
$ time ../meta-scripts/sources.sh --cache-file=sources-copy.json > sources.json

real    1m15.459s
user    1m14.891s
sys     0m0.324s
$ # after
$ time ../meta-scripts/sources.sh --cache-file=sources-copy.json > sources.json

real    0m15.600s
user    0m15.071s
sys     0m0.477s

$ # with `sources.json` being correctly generated and unchanged from the checkout.
```
After tianon's improvements and a reorder 😍 
```console
$ time ../meta-scripts/sources.sh --cache-file=sources-copy.json > sources.json

real    0m5.095s
user    0m4.570s
sys     0m0.277s
```

And on a Jenkins worker versus the ~7 minutes in #115: 🎉 🎊 🥳 
```console
$ time ../doi-meta-scripts/sources.sh --cache-file=sources-copy.json > sources-cached.json

real    0m44.134s
user    0m43.897s
sys     0m1.458s
```

### DIFF
--- a/sources.sh
+++ b/sources.sh
@@ -219,7 +219,42 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
 		# https://unix.stackexchange.com/a/738744/153467
 		reduce .[] as $a ([]; if IN(.[]; $a) then . else . += [$a] end)
 	;
-	reduce .[] as $in ({};
+	def meld($o):
+		# recursive merge of objects like "*", but also append lists (uniquely) instead of replace
+		# https://stackoverflow.com/a/53666584, but with lists unique-ified and less work being done
+		if type == "object" and ($o | type) == "object" then
+			reduce ($o | keys_unsorted[]) as $k (.;
+				.[$k] |= meld($o[$k])
+			)
+		elif type == "array" and ($o | type) == "array" then
+			. + $o
+			| unique_unsorted
+		elif $o == null then
+			.
+		else
+			$o
+		end
+	;
+	(
+		# creating a lookup of .[tag][arch] to a list of sourceIds
+		[
+			.[] as $s
+			| ( $s.arches[] | .tags[], .archTags[] )
+			| {
+				key: .,
+				# this happens pre-arches-merge, so it is only one arch
+				value: { ($s.arches | keys[]): [$s.sourceId] },
+			}
+		]
+		# do not try to code golf this to one reduce without from_entries or group_by; doing ".[xxx] |=" on the *full* unordered object gets orders of magnitude slower
+		| group_by(.key)
+		| [
+			# many little reduces based on same key (group_by^), instead of one very big and expensive
+			.[] | reduce .[] as $r ({}; meld($r))
+		]
+		| from_entries
+	) as $tagArches
+	| reduce .[] as $in ({};
 		.[$in.sourceId] |=
 			if . == null then
 				$in
@@ -253,17 +288,6 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
 	)
 	# TODO a lot of this could be removed/parsed during the above reduce, since it has to parse things in build order anyhow
 	# TODO actually, instead, this bit should be a totally separate script so the use case of "combine sources.json files together" works better 👀
-	| (
-		# TODO make this faster, this reduce takes the longest time now
-		reduce to_entries[] as $e ({};
-			$e.key as $sourceId
-			| .[ $e.value.arches[] | .tags[], .archTags[] ] |= (
-				.[$e.value.arches | keys[]] |= (
-					. + [$sourceId] | unique_unsorted
-				)
-			)
-		)
-	) as $tagArches
 	| map_values(
 		.arches |= with_entries(
 			.key as $arch


sources.sh
```diff
@@ -219,6 +219,22 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
 		# https://unix.stackexchange.com/a/738744/153467
 		reduce .[] as $a ([]; if IN(.[]; $a) then . else . += [$a] end)
 	;
+	def meld($a; $b):
+		# recursive merge of objects like "*", but also append lists (uniquely) instead of replace
+		# https://stackoverflow.com/a/53666584, but with lists unique-ified
+		if ($a | type) == "object" and ($b | type) == "object" then
+			reduce ([$a, $b] | add | keys_unsorted[]) as $k ({};
+				.[$k] = meld( $a[$k]; $b[$k])
+			)
```
tianon, line 228: This is doing a lot of work twice :joy:

```suggestion
			reduce ([ $a, $b | keys_unsorted[] ] | unique_unsorted[]) as $k ({};
				.[$k] = meld($a[$k]; $b[$k])
			)
```

sources.sh
```diff
@@ -219,6 +219,22 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
+		elif ($a | type) == "array" and ($b | type) == "array" then
+			$a + $b
+			| unique_unsorted
+		elif $b == null then
+			$a
+		else
+			$b
+		end
```
tianon, line 236: Even better, ditch `$a` and use `.` and everything simplifies:

```suggestion
	def meld($o):
		# recursive merge of objects like "*", but also append lists (uniquely) instead of replace
		# https://stackoverflow.com/a/53666584, but with lists unique-ified and less work being done
		if type == "object" and ($o | type) == "object" then
			reduce ($o | keys_unsorted[]) as $k (.;
				.[$k] |= meld($o[$k])
			)
		elif type == "array" and ($o | type) == "array" then
			. + $o
			| unique_unsorted
		elif $o == null then
			.
		else
			$o
		end
```

(untested)

sources.sh
```diff
@@ -254,15 +270,26 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
-				)
+		# creating a lookup of .[tag][arch] to a list of sourceIds
+		[
+			.[] as $e
+			| ( $e.arches[] | .tags[], .archTags[] )
+			| {
+				key: .,
+				value: {
+					($e.arches | keys[]): [$e.sourceId]
+				}
```
tianon, line 281:
```suggestion
				value: { ($e.arches | keys[]): [$e.sourceId] },
```

sources.sh
```diff
@@ -254,15 +270,26 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
-		# TODO make this faster, this reduce takes the longest time now
-		reduce to_entries[] as $e ({};
-			$e.key as $sourceId
-			| .[ $e.value.arches[] | .tags[], .archTags[] ] |= (
-				.[$e.value.arches | keys[]] |= (
-					. + [$sourceId] | unique_unsorted
-				)
+		# creating a lookup of .[tag][arch] to a list of sourceIds
+		[
+			.[] as $e
```
tianon, line 275: Can we replace `$e` with `$s`?  When I see `$e` my brain says "entry" but this is a "source" object :innocent:

sources.sh
```diff
@@ -254,15 +270,26 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
+			| ( $e.arches[] | .tags[], .archTags[] )
+			| {
+				key: .,
+				value: {
+					($e.arches | keys[]): [$e.sourceId]
```
tianon, line 280: This is _technically_ looping over arches twice and creating combinations that don't exist (just like the previous implementation), but it's fine.

sources.sh
```diff
@@ -254,15 +270,26 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
+				}
+			}
+		]
+		# do not try to code golf this to one reduce without from_entries or group_by, that is very slow
```
tianon, line 284:
```suggestion
		# do not try to code golf this to one reduce without from_entries or group_by; doing ".[xxx] |=" on the *full* unordered object gets orders of magnitude slower
```

sources.sh
yosifkit: So... yeah, moving this to before the reduce in order to get rid of that duplication (along with your meld improvements), makes it even faster!!!

```console
$ time ../meta-scripts/sources.sh --cache-file=sources-copy.json > sources.json

real    0m5.095s
user    0m4.570s
sys     0m0.277s
```

tianon (dismissed):

sources.sh
```diff
@@ -219,7 +219,41 @@ jq <<<"$sources" --argjson pins "$externalPinsJson" '
...
+		end
+	;
+	(
+		# creating a lookup of .[tag][arch] to a list of sourceIds
+		[
+			.[] as $s
+			| ( $s.arches[] | .tags[], .archTags[] )
+			| {
+				key: .,
+				value: { ($s.arches | keys[]): [$s.sourceId] }
```
tianon, line 245:
```suggestion
				value: { ($s.arches | keys[]): [$s.sourceId] },
```

tianon: Maybe we should also add a comment that this happens pre-arches-merge so looping over arches twice is fine?

tianon (approved):

PR #119, opened 2025-03-03
Update deploy to have output during dry-run too

tianon:
This moves the "normal" deploy output from stdout to stderr so that we can print it all the time (whether it's a dry-run or not), such that dry-run can have the same progress output.

~~For now, I've decided that "needs-push" means "failure" and printed the status accordingly, but I could be convinced it should do something else instead.~~ (https://github.com/docker-library/meta-scripts/pull/119#discussion_r1978375608)

Without this, "dry-run" doesn't print any status, especially in the case of 100% "already pushed" (which is the whole reason we implemented dry-run in the first place, so it'd helpful for it to have *some* output).

### DIFF
--- a/cmd/deploy/main.go
+++ b/cmd/deploy/main.go
@@ -203,6 +203,9 @@ func main() {
 					defer lock.(*sync.RWMutex).RUnlock()
 				}
 
+				logText := ref.StringWithKnownDigest(refsDigest) + logSuffix
+				fmt.Fprintln(os.Stderr, startedPrefix+logText)
+
 				if dryRun {
 					needsDeploy, err := normal.dryRun(ctx, ref)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/deploy/main.go
```diff
@@ -217,10 +220,13 @@ func main() {
 							panic(err) // TODO exit in a more clean way (we can't use "os.Exit" because that causes *more* errors 😭)
 						}
 						dryRunOut <- j
+
+						// TODO this isn't exactly *failure* -- maybe we should just add a logText suffix instead?
+						fmt.Fprintln(os.Stderr, failurePrefix+logText)
```
tianon, line 225: A good argument in favor of dropping the `TODO` here and just accepting that this *is* failure is that this case is basically "we do need to push, but were blocked by configuration"

yosifkit: I agree that an unpushed image should be considered a failure on `dry-run`. Should this also exit at the end with an error if there were things needing to be pushed when doing `dry-run`?

tianon: That's a harder question -- I think it's more useful/easier to use if "non-empty stdout" means something needed to be pushed, since the whole point is querying what _did_ need to be pushed, and a non-zero exit code then makes it harder to differentiate breaking errors vs just "things need to be pushed".

yosifkit (approved):

PR #120, opened 2025-03-03
Fix `deploy --dry-run` deadlock

tianon:
Once we got over 1000 items in our deploy list, we deadlocked because nothing consumes the channel until we're done with the "queue everything up" loop, so we need to start consuming that list right away instead.  As long as it's part of our waitgroup, this should be fine and work correctly.


tianon:
For a confidence boost, we _do_ log this output, so we know that it's still working correctly, and still outputting in-order.

yosifkit (approved):

PR #123, opened 2025-03-19
Improve queue ordering by waiting a moment between triggering

yosifkit:
> using "waitForStart" means we have to wait for the entire "quietPeriod" before we get to move on and schedule more


tianon (approved):

tianon:
As a note for (our) posterity, @yosifkit also tried values smaller than 1 here (0.5) and it doesn't accept them, so we'd have to use `sleep` instead if we want a limit smaller than this.

PR #124, opened 2025-04-25
Move `oci-import` (and SBOM) shell code out of `meta.jq` into an explicit shell script

tianon:
Also, add a *lot* more validation.

This new script/code assumes we're setting `BASHBREW_META_SCRIPTS` to an appropriate value (pointing to the directory of this repository on-disk).


tianon:
(I made https://github.com/tianon/dockerfiles/pull/982 so I could actually test this in a real load-bearing way)

oci.jq
```diff
@@ -88,52 +97,190 @@ def sort_manifests:
...
+	# TODO artifactType?
+
+	# https://github.com/opencontainers/image-spec/blob/v1.1.1/image-index.md#image-index-property-descriptions
+	| if has("platform") then
+		validate(.platform;
+			validate_IN(type; "object")
+			| validate(.architecture; type == "string" and length > 0)
+			| validate(.os; type == "string" and length > 0)
+			| if has("os.string") then
+				validate(."os.string"; type == "string" and length > 0)
```
yosifkit, line 217: Did you mean `os.version` here?

yosifkit (approved):

PR #125, opened 2025-04-30
Implement more TODOs in `oci-validate` code

tianon:
I'm much happier with this interface now -- it can properly handle an image with attestations now, for example.

### DIFF
--- a/helpers/oci-validate.sh
+++ b/helpers/oci-validate.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-# given an OCI image layout (https://github.com/opencontainers/image-spec/blob/v1.1.1/image-layout.md), verifies all descriptors as much as possible (digest matches content, size, some media types, layer diff_ids, etc)
+# given an OCI image layout (https://github.com/opencontainers/image-spec/blob/v1.1.1/image-layout.md), verifies all descriptors as much as possible (digest matches content, size, media types, layer diff_ids, etc)
 
 layout="$1"; shift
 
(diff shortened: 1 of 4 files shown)


tianon:
I think I'm happy enough with this that it could become part of our builds, actually. :thinking:

Longer-term I want to munge _all_ our builds (even classic builder) to output an OCI layout, so this could become a standard part of the pipeline.  For that we would want to get even more specific (`index.json` has only a single [index] entry, that index has only a single image and optionally a single attestation).

tianon:
```console
invalid .["manifests"][0]:
	{"mediaType":"application/vnd.oci.image.index.v1+json","digest":"sha256:9DEE8c8868c2b379281f0956df25a86acb830c1f3eb88aa23a5a710547fe3165","size":1298}
invalid .["digest"]:
	"sha256:9DEE8c8868c2b379281f0956df25a86acb830c1f3eb88aa23a5a710547fe3165"
invalid value:
	{"algorithm":"sha256","encoded":"9DEE8c8868c2b379281f0956df25a86acb830c1f3eb88aa23a5a710547fe3165"}
invalid .["encoded"]:
	"9DEE8c8868c2b379281f0956df25a86acb830c1f3eb88aa23a5a710547fe3165"
the encoded portion MUST match /[a-f0-9]{64}/
```

"this brings joy"

helpers/oci-validate.sh
```diff
@@ -23,85 +23,131 @@ jq -L"$BASHBREW_META_SCRIPTS" --slurp '
...
+	local digest size dataDigest= dataSize=
+	digest="$("${algo}sum" "$file" | cut -d' ' -f1)"
+	digest="$algo:$digest"
+	size="$(stat --dereference --format '%s' "$file")"
+	if [ "$data" != ' ' ]; then
+		dataDigest="$(base64 <<<"$data" -d | "${algo}sum" | cut -d' ' -f1)"
+		dataDigest="$algo:$dataDigest"
+		dataSize="$(base64 <<<"$data" -d | wc --bytes)"
+		# TODO *technically* we could get clever here and pass `base64 -d` to something like `tee >(wc --bytes) >(dig="$(sha256sum | cut -d' ' -f1)" && echo "sha256:$dig" && false) > /dev/null` to avoid parsing the base64 twice, but then failure cases are less likely to be caught, so it's safer to simply redecode (and we can't decode into a variable because this might be binary data *and* bash will do newline munging in both directions)
+	fi
```
yosifkit, line 63: Should this be able to handle a descriptor that only has `data` and not also a file on disk? Or is that not a valid OCI layout?

tianon: This is such an (unintentionally) brutal comment :joy: :sob:

You're absolutely right, and making me question whether I like this interface again.

Implementations worried about portability should absolutely have both `data` _and_ the blob in the `blobs/` folder, but it's not technically invalid not to, and I'd love to support that. :thinking:

yosifkit (approved):
LGTM, we can add more extensive `data` field using later since it isn't necessary right now.

yosifkit (approved):
Still looks good to me.

PR #128, opened 2025-08-18
oci: do not require top-level mediaType

cyphar:
The top-level mediaType member was added in response to CVE-2021-41190,
but while it is suggested (SHOULD) it is not required (MUST) and some
older tools do not fill this mediaType field (such as skopeo, at least
for "index.json").

I plan to use these jq-based validation scripts for umoci, but
incompatibility with skopeo is a little annoying (since that is what we
use to pull images for our tests). We can work around it for
"index.json", but it seems incorrect to claim that an image is invalid
because of a missing suggested field.

Instead, add an informational message but still permit such images.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/oci.jq
+++ b/oci.jq
@@ -250,11 +250,28 @@ def validate_oci_subject_haver:
 	else . end
 ;
 
+# Some objects have .mediaType fields which SHOULD NOT be unset but are still
+# valid if they are missing, so we should emit a warning if they are unset but
+# do not error out.
+# usage: same as validate_IN(.mediaType; options).
+# https://github.com/opencontainers/image-spec/security/advisories/GHSA-77vh-xpmg-72qh (CVE-2021-41190)
+# https://github.com/opencontainers/image-spec/blob/v1.1.1/manifest.md
+# https://github.com/opencontainers/image-spec/blob/v1.1.1/image-index.md
+def validate_optional_mediatype(options):
+	if has("mediaType") then
+		validate_IN(.mediaType; options)
+	else [
+		# Output a warning message.
+		("warning: top-level mediaType field is missing from object (see CVE-2021-41190)\nexpected one of:\n\t\([ options | tojson ] | join("\n\t"))\n" | stderr | empty),
+		.
+	] | last end
+;
+
 # https://github.com/opencontainers/image-spec/blob/v1.1.1/image-index.md
 def validate_oci_index($opt):
 	validate_IN(type; "object")
 	| validate_IN(.schemaVersion; 2)
-	| validate_IN(.mediaType; media_types_index)
+	| validate_optional_mediatype(media_types_index)
 	| if has("artifactType") then
 		validate(.artifactType; type == "string")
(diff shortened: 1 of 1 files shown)


polarathene:
> I've toyed with the idea of doing something with [`gojq`](https://github.com/itchyny/gojq) (where I could inject my own helper functions), but the lack of order-preserving maps is a complete non-starter for me.

I'm not 100% sure if `jaq` doesn't have that problem (_[this docs entry suggests it doesn't](https://gedenkt.at/jaq/manual/#--sort-keys)_), but the other concern with base64 that you link to on the cryptographic hash issue comment isn't a problem with `jaq` (_but binary data does seem reliant upon an upcoming 3.0 release, so you'd need to wait on that_).

`jaq` has a library available in addition to the CLI, so if you're comfortable with Rust it might be an option that works for where `gojq` wasn't.

`jaq` also has `import` so that you can [load in another JSON file](https://gedenkt.at/jaq/manual/#data-import), but after trying that I've realized it [wouldn't work in this situation](https://github.com/01mf02/jaq/issues/373).

The assert issue linked can be adapted to `try/error/catch` with `jaq`? (_**EDIT:** Nevermind this is the [same syntax with `jq`](https://jqlang.org/manual/#error) already, I misunderstood the PR_):

```console
$ jaq -n '[{a: 4, b: 2}, {a: 7, b: 6}] | try error(map(.b) | add > 10) catch("your message here")'
"your message here"
```

---

`nushell` is an entire shell, but it has comparison pages to `jq`. Using it's own DSL to do the same operations technically matches your requirements to not call out to individual CLI tools. It does have the operations you're interested in:
- Assert capabilities: https://www.nushell.sh/book/testing.html#assert-commands
- SHA256: https://www.nushell.sh/commands/docs/hash_sha256.html
- Base64 encode/decode (_and [other format/encoding conversions](https://www.nushell.sh/commands/categories/formats.html)_): https://www.nushell.sh/commands/docs/encode_base64.html
- Open/Load files such as JSON: https://www.nushell.sh/commands/docs/open.html

You can use various operators and pipes with nushell, along with define custom methods, etc. It feels a bit awkward as a shell sometimes, but fairly capable of processing data. It does come with the convenience of a [container](https://github.com/nushell/nushell/pkgs/container/nushell) or [static binary](https://github.com/nushell/nushell/releases/tag/0.108.0) but it's also a bit heavy in size for a utility compared to `jq`.

There's also `yq` which is fairly similar to `jq` but does have a `load` operator you need, and it has a [Go library package](https://github.com/mikefarah/yq/tree/master/pkg/yqlib).

`fq` (also Go) is a `jq` alternative with a focus on binary support. It has a [`tosha256` operator](https://github.com/wader/fq/blob/90d2e5e43fd2cceeebdf5ea291b1552f21bf6130/doc/usage.md?plain=1#L878) (_**EDIT:** I see [it was already mentioned at the jq issue](https://github.com/jqlang/jq/issues/1116#issuecomment-1315190648)_), along with [`open`](https://github.com/wader/fq/blob/90d2e5e43fd2cceeebdf5ea291b1552f21bf6130/doc/usage.md?plain=1#L598) to read in a file.

PR #130, opened 2025-09-08
Move windows builds to auto scaling groups

yosifkit:
Now that we have auto scaling groups, let's move more off of GHA. Since we don't have virtualization extensions, the build job needs the Window OS version to select the right node labels so that the correct autoscaling group is used (OS version must match or it can't run/build the Docker image).

One of the benefits is that with our Trigger jobs, we can actively wait on the queue of Build jobs in Jenkins instead of the "throw it over the wall" of GHA.

We'll also need to enable the [Windows build job](https://doi-janky.infosiftr.net/job/meta/job/windows-amd64/job/build/) after merging.

### DIFF
--- a/jenkins.jq
+++ b/jenkins.jq
@@ -20,6 +20,26 @@ def crane_deploy_commands:
 ;
 
 # input: "build" object (with "buildId" top level key)
+# output: json object with windowsVersion: "2022" or "2025" or empty {} if there is no os.version like for Linux
+def windows_version:
+	[ .build.resolvedParents[]?.manifests[]?.platform? | select(has("os.version")) | ."os.version" ][0] // ""
+	| if . != "" then
+		{ windowsVersion: (
+			# https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle
+			# https://github.com/microsoft/hcsshim/blob/d9a4231b9d7a03dffdabb6019318fc43eb6ba996/osversion/windowsbuilds.go
+			capture("^10[.]0[.](?<build>[0-9]+)([.]|$)")
+			| {
+				# since this is specifically for GitHub Actions support, this is limited to the underlying versions they actually support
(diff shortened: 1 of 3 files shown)


jenkins.jq
```diff
@@ -19,6 +19,27 @@ def crane_deploy_commands:
...
+			# https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle
+			# https://github.com/microsoft/hcsshim/blob/d9a4231b9d7a03dffdabb6019318fc43eb6ba996/osversion/windowsbuilds.go
+			capture("^10[.]0[.](?<build>[0-9]+)([.]|$)")
+			| {
+				# since this is specifically for GitHub Actions support, this is limited to the underlying versions they actually support
+				# https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+				"26100": "2025", # https://oci.dag.dev/?image=mcr.microsoft.com/windows/servercore:ltsc2025
+				"20348": "2022", # https://oci.dag.dev/?image=mcr.microsoft.com/windows/servercore:ltsc2022
+				"17763": "2019", # https://oci.dag.dev/?image=mcr.microsoft.com/windows/servercore:ltsc2019
+				"": "",
```
tianon, line 37: Hmm, I don't remember why we added this, but we should probably drop it and let it fall back to `unknown` instead.

```suggestion
```

tianon (approved):

jenkins.jq
tianon: Yeah, I spent a lot of time debating that same thing, but decided the simple change is fine for now and we can improve later. :joy:

tianon: (otherwise every call site has to be more complicated, and I didn't love that, but maybe it's correct - this seems fine for now, especially as it's scoped to just `jenkins.jq` :+1:)

PR #134, opened 2025-09-13
Move amd64 and i386 builds to auto scaling groups

yosifkit:
Following https://github.com/docker-library/meta-scripts/pull/130, this moves the Linux `amd64` and `i386` builds to our auto scaling groups as well.

There are now unused functions (like [`gha_payload`](https://github.com/docker-library/meta-scripts/blob/86288421105378da0d6543766e55ede9c5f60cf6/jenkins.jq#L44-L55)) to remove once we no longer want to come back to GHA for building.

We'll also need to enable the [build](https://doi-janky.infosiftr.net/job/meta/job/amd64/job/build/) [jobs](https://doi-janky.infosiftr.net/job/meta/job/i386/job/build/) after merging.

### DIFF
--- a/jenkins.jq
+++ b/jenkins.jq
@@ -62,13 +62,8 @@ def get_arch_queue($arch):
 			needs_build
 			and .build.arch == $arch
 		)
-		|  if IN(.build.arch; "amd64", "i386") then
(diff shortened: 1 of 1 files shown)


tianon (approved):

tianon:
Maybe I merge this on Monday, not right before the weekend. :joy:

PR #146, opened 2026-08-18
Support limited cross-buildability

tianon:
Specifically, this requires `doi.jq`'s new `build_can_cross` function to return `true` (which it currently does *only* for `oci-import` images), *and* a Jenkins node with the `cross-ARCH` label *and* a credential for the arch with a `-cross` suffix.

The build is also targeted such that if a native node is available, it will satisfy the label expression and be used (which is what allows this to continue working as-is without any infra changes and not perform any "cross" builds).

Assisted-By: "claude my eyes right out"


tianon:
The biggest "gap" here for cross-building anything besides `oci-import` is that `bashbrew` will neither parse nor allow `FROM --platform` so we can't *actually* do successful cross-builds with this without *also* having something like https://github.com/docker-library/bashbrew/commit/c2bfb6ae301212def0112633afa15714a4f72abd *and* solving the edge cases it points at.

